PR #36, opened 2014-11-23
Update info about superscripts and subscripts

srpearce:
@mhoeber @lamagnifica @catong Please review. If it looks good, go ahead and merge it - I'll be in transit most of the day on Monday so I won't be able to check in until late afternoon/early evening.

### DIFF
--- a/en_us/students/source/SFD_mathformatting.rst
+++ b/en_us/students/source/SFD_mathformatting.rst
@@ -56,18 +56,28 @@ math programs before.
 Entering Math Expressions
 ****************************
 
-When you enter your plain text into the calculator or the response field, follow
-these guidelines:
+When you enter your plain text into the calculator or the response field,
+follow these guidelines:
 
-* Use standard arithmetic operation symbols: the plus sign (+), minus sign (-),
-  multiplication sign (*), and division sign (/).
-* Indicate multiplication explicitly. That is, instead of ``mc^2`` type
-  ``m*c^2``, and instead of ``5a+4b+3c`` type ``5*a+4*b+3*c``.
-* Use parentheses to specify the order of operations and to make your expression
-  as clear as possible. However, only use curved parentheses (``()``). Don't use
-  brackets (``[]``) or braces (``{}``).
-* Use a caret (^) to indicate an exponent.
-* Use an underscore (_) to indicate a subscript.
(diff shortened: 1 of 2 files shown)


en_us/students/source/SFD_mathformatting.rst
```diff
@@ -66,8 +66,14 @@ these guidelines:
...
-* Use a caret (^) to indicate an exponent.
-* Use an underscore (_) to indicate a subscript.
+* Use a caret (^) to indicate an exponent or superscript. If the exponent or
+  superscript has multiple characters or is an expression, surround the
+  expression with parentheses. For example, you can enter ``x^n`` or
+  ``x^(n-1)``.
+* Use an underscore (_) to indicate a subscript. If the subscript has multiple
+  characters, type the characters without a space. For example, you can enter
+  ``v_IN-v_OUT``. Note, however, that subscripts cannot currently include
+  operators or parentheses.
```
lamagnifica: if there is time, I'd prefer to see these bullets reformatted to put the subject first. For example,
- **Exponents or superscripts**: Insert a caret (^) before an exponent or superscript. If the exponent or superscript includes multiple characters or is an expression, surround the expression with parentheses. For example, enter `x^n` or `x^(n-1)`.
- **Subscripts**: Insert an underscore (_) before a subscript...

I can do this quickly today if @mhoeber and @catong think it would be worthwhile.

mhoeber:
@lamagnifica  I made change you suggested.  thumbs up?

lamagnifica:
@mhoeber I meant the full list, not just these two bullets (and I meant that I would do it!). Here are the bullets from the beginning of the list (the ones at the end of the list are ok as is:
- **Arithmetric operations**: Use standard characters for addition (+), subtraction (-),
  multiplication (*), and division (/).
  - **Multiplication**: Be sure to indicate multiplication explicitly. That is, instead of `mc^2` type
    `m*c^2`, and instead of `5a+4b+3c` type `5*a+4*b+3*c`.
- **Operation Order**: Use parentheses (`( )`) to specify the order of operations and to make your  expression as clear as possible. Use curved parentheses (`( )`) only. Do not use brackets (`[ ]`) or braces (`{ }`).

lamagnifica:
:+1:

mhoeber:
@lamagnifica  thanks changes made

PR #37, opened 2014-11-25
Fix html sample and note that full tags needed

catong:
@adampalay @ahodges can you please review
FYI @mhoeber


adampalay:
:+1:

mhoeber:
:+1:

PR #46, opened 2014-12-03
Sylvia/docs/doc 1132

srpearce:
@mhoeber @lamagnifica @catong Please review. Thanks!


mhoeber:
one note:  is possible that instructor hides wiki for class -- make it clear that students may not see it at all.

otherwise looks good. :+1:

PR #56, opened 2014-12-10
Clytwynec/add travis configuration

clytwynec:
@benpatterson Please review. 

For now, this runs only the html builds (no latexpdf).  I've made it so that builds will fail if there are sphinx warnings during the build, even if the build itself doesn't error out.  @mhoeber Is that the expected behavior?

Most recent build: https://travis-ci.org/clytwynec/edx-documentation/builds/43651494


run_tests.sh
```diff
@@ -0,0 +1,50 @@
...
+    err_log_file=$err_log_dir/err.log
+    mkdir -p $err_log_dir
+    
+    # Generate html docs.
+    # -w writes warnings and errors to the specified file
+    # in addition to stderr.
+    make html SPHINXOPTS="-w $err_log_file"
+    
+    # Get errors from the error log
+    IFS=$'\n' read -d '' -r -a errors < $err_log_file
```
benpatterson: Just confirming...we're looking for errors only, no warnings?

run_tests.sh
```diff
@@ -0,0 +1,50 @@
...
+    
+    if [ $num_errors -gt 0 ]; then
+        BUILD_STATUS=1
+        TOTAL_BUILD_ERRORS=$((TOTAL_BUILD_ERRORS + num_errors))
+    fi
+done
+
+echo '********** All builds done ************'
+echo SPHINX ERRORS AND WARNINGS: $TOTAL_BUILD_ERRORS
+echo BUILD EXIT STATUS: $BUILD_STATUS
```
benpatterson: How about an if/then here, to echo "build failed" or something a bit more obvious to non-bashy folks, and a "build passed" in the event of a success ?

Also, if we're doing errors-only, should we capture warnings and print them when giving the final status? Might be useful to know the number of warnings, even if we don't fail based on them.

run_tests.sh
```diff
@@ -0,0 +1,50 @@
...
+    # -w writes warnings and errors to the specified file
+    # in addition to stderr.
+    make html SPHINXOPTS="-w $err_log_file"
+    
+    # Get errors from the error log
+    IFS=$'\n' read -d '' -r -a errors < $err_log_file
+
+    num_errors=${#errors[@]}
+    echo SPHINX ERRORS: $num_errors
+    echo
```
benpatterson, line 69: Assuming this is a way to get a newline....just thought I'd point it out though in case this was a duplicate echo command.

run_tests.sh
```diff
@@ -0,0 +1,50 @@
...
+    # Generate html docs.
+    # -w writes warnings and errors to the specified file
+    # in addition to stderr.
+    make html SPHINXOPTS="-w $err_log_file"
+    
+    # Get errors from the error log
+    IFS=$'\n' read -d '' -r -a errors < $err_log_file
+
+    num_errors=${#errors[@]}
+    echo SPHINX ERRORS: $num_errors
```
benpatterson, line 67: Commented on this below, but this might be a good place to echo the number of warnings.

mhoeber:
@clytwynec   currently we complete builds with warnings.  we want to get rid of warnings but we're not there yet.  So can this complete with warnings?

run_tests.sh
clytwynec: This is actually looking for warnings too.

clytwynec:
@mhoeber As is, this will complete each of the html builds even if there are warnings, but it will mark the travis build as a whole as failing.  You are saying that  it should be considered passing even with the sphinx warnings as long as the html docs are able to be built, correct?

clytwynec:
@benpatterson  I've updated with your suggestions. Here's the new build: https://travis-ci.org/clytwynec/edx-documentation/builds/43746482

benpatterson:
LGTM :+1:

clytwynec:
@singingwolfboy Any chance you can enable travis builds for this repo?

singingwolfboy:
@clytwynec Travis should now be enabled for this repo.

clytwynec:
Thanks @singingwolfboy !

PR #64, opened 2014-12-15
Missed edits from Rachel

lamagnifica:
@rjgrubb, thanks for catching this oversight. 
@catong, would you be so kind as to check the changed lines to make sure I did not introduce any problems. (Today if possible.)


catong:
:+1:

PR #45, opened 2014-12-03
Assign students to cohorts by uploading a csv file

catong:
First draft of doc: course authors can use a csv upload to assign students to cohorts.
Will continue to update if additional info surfaces about UTF-8 and other requirements/limitations.

PR #65, opened 2014-12-15
Remove usernames as Unicode use example

catong:
Amendment to https://github.com/edx/edx-documentation/pull/45
Turns out we do not support Unicode characters in usernames, though we still require UTF-8 encoded CSV files. This change removes "for example, usernames" in the "Creating a Unicode-encoded CSV File" topic so that we don't seem to imply that we support usernames with Unicode characters.
@explorerleslie @dan-f @lamagnifica please give thumbs if OK. I'd like to merge this today. Thanks!
FYI @WatsonEmily


dan-f:
:+1:

explorerleslie:
:+1:

lamagnifica:
:+1:

catong:
Merged https://github.com/edx/edx-documentation/pull/65

PR #72, opened 2014-12-18
Ahodges/doc1278

lamagnifica:
@stroilova, please review. On your ok, I will merge with other changes from mobile.


lamagnifica:
This was only intended to add the "code" field to seek_video and stop_video, as described in DOC-1278. Live docs do have this change.
All files saved locally in case of future usefulness.

PR #84, opened 2014-12-30
CSV upload: remove column name case requirement

catong:


catong:
Merging https://github.com/edx/edx-documentation/pull/84

PR #85, opened 2014-12-30
Fix punctuation for file type

catong:
Replace comma with period


catong:
Merging https://github.com/edx/edx-documentation/pull/85

PR #114, opened 2015-01-21
Fixes a broken link

lamagnifica:
@catong would you please take a look?
this is where it should be pointing... http://draft-processing-video-files.readthedocs.org/en/latest/video_uploads.html#reporting-video-statuses


catong:
Built locally with that link and confirmed it works correctly. :+1:

lamagnifica:
thanks!

PR #116, opened 2015-01-22
Install-Config Guide change log and theme fix

mhoeber:
@lamagnifica  just quickly fixed change log from recent changes and applied edx theme (you can ignore conf.py file)


lamagnifica:
:+1:

PR #125, opened 2015-01-27
Developers guide - workaround for failing redirects

mhoeber:
@sarina  these files with includes will create those files at URL without the subfolder.  Updated image paths to be absolute for project so they'll work in both pages.


sarina:
:+1:

PR #128, opened 2015-01-28
Remove requirements doc to make RTD work

mhoeber:
@mulby   I don't fully understand why; having the top level requirements file is breaking RTD.  I'm removing for now, updated readme to point to travis requirements.

FY @lamagnifica @srpearce @catong


mulby:
odd, sorry for breaking things! maybe RTD doesn't like the fact that it's a symlink?

PR #150, opened 2015-02-12
Fixes DOC-1672

lamagnifica:
@mulby please review for accuracy
@benpatterson FYI
@srpearce please take a look

### DIFF
--- a/en_us/data/source/internal_data_formats/tracking_logs.rst
+++ b/en_us/data/source/internal_data_formats/tracking_logs.rst
@@ -348,8 +348,6 @@ events that originate on the server (during the processing of a request).
 Enrollment Events
 =========================
 
-.. tracked_command.py
-
 ``edx.course.enrollment.activated`` and ``edx.course.enrollment.deactivated``
 *****************************************************************************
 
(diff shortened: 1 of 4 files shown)


en_us/data/source/internal_data_formats/tracking_logs.rst
```diff
@@ -429,7 +427,37 @@ Example
...
+   * - Field
+     - Type
+     - Details
+   * - ``course_id``
+     - string
+     - See the description of the ``context.course_id`` field in
+       :ref:`context`.
+   * - ``mode``
+     - string
+     - Identifies the mode that the new enrollment type for the student.
```
srpearce: that --> of

srpearce:
One typo, otherwise :+1:

en_us/data/source/internal_data_formats/tracking_logs.rst
```diff
@@ -429,7 +427,37 @@ Example
...
+   :widths: 15 15 60
+   :header-rows: 1
+
+   * - Field
+     - Type
+     - Details
+   * - ``course_id``
+     - string
+     - See the description of the ``context.course_id`` field in
+       :ref:`context`.
```
mulby: This isn't exactly correct. The value in the context represents the course that the user is currently viewing. The value here represents the course that is being acted upon. There are some tools, for example that allow you to enroll users without actually viewing a webpage. In that case the value in the context would be null and this value would contain the course that the user was enrolled in.

Hopefully that makes sense. Also the history of this field as called out in the activated/deactivated event description is not correct, it is not here just for historical reasons - it is here because of the reason described above.

en_us/data/source/internal_data_formats/tracking_logs.rst
```diff
@@ -429,7 +427,37 @@ Example
...
+   * - ``mode``
+     - string
+     - Identifies the mode that the new enrollment type for the student.
+       'audit', 'honor', 'verified'.
```
mulby: and professional?

en_us/data/source/internal_data_formats/tracking_logs.rst
```diff
@@ -429,7 +427,37 @@ Example
...
+   * - ``user_id``
+     - integer
+     - Identifies the student.
```
mulby: It might be worth calling out here that this may be _different_ than the user_id in the context / the username at the root of the event.

If, for example, an instructor bulk enrolls a bunch of students in a course, then the username at the root and the user_id in the context will be the instructor's user information, however, this user_id in the event payload will be the user that is being enrolled in the course.

mulby:
:+1:

PR #154, opened 2015-02-13
Fixes DOC-1676

srpearce:


srpearce:
@mhoeber @lamagnifica @catong Can one of you give me quick thumbs? I just discovered a large errant image in our Saudi docs for course prerequisites, and since prereqs are fairly high-profile I thought we should probably fix it quickly.

lamagnifica:
:+1:

PR #170, opened 2015-02-19
Fixes DOC-955

lamagnifica:
Wording change requested by @stroilova

### DIFF
--- a/en_us/data/source/internal_data_formats/credentials.rst
+++ b/en_us/data/source/internal_data_formats/credentials.rst
@@ -221,9 +221,19 @@ Browser. Alternatively, you can use the `AWS Command Line Interface`_.
    that account. For more information, refer to the documentation for the tool
    that you selected.
 
-   Data package files are in the edX **course-data** and
-   **edx-course-data** buckets. For information about the files that you
-   download from Amazon S3, see :ref:`Package`.
+#. Select the **s3://course-data** bucket to access the database data files. 
+   
+   Select the **s3://edx-course-data/{org}** folder to access the event data
+   files. You must include the identifier for your organization after the
+   name of the bucket.
+
+   .. note:: If you are using a third-party tool to connect to Amazon S3, you 
+    may not be able to navigate directly between the **s3://course-data**
+    bucket and the **s3://edx-course-data/{org}** folder. You may need to
+    disconnect from Amazon S3 and then reconnect to the other destination.
+
+For information about the files in each Amazon S3 destination, see
+:ref:`Package`.
 
 .. _AWS Command Line Interface: http://aws.amazon.com/cli/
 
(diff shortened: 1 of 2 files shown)


en_us/data/source/internal_data_formats/credentials.rst
```diff
@@ -221,9 +221,9 @@ Browser. Alternatively, you can use the `AWS Command Line Interface`_.
    that account. For more information, refer to the documentation for the tool
    that you selected.
 
-   Data package files are in the edX **course-data** and
-   **edx-course-data** buckets. For information about the files that you
-   download from Amazon S3, see :ref:`Package`.
+   Database data files are in the **s3://course-data** bucket. Event data
+   files are in the **s3://edx-course-data/{org}** bucket. For information
```
stroilova: This is great @lamagnifica ! Do you think we could add a comment here that the AWS S3 connection strategies are different for these two paths?
Specifically, the data czar will have to specify the full path to their s3://edx-course-data/{org} folder explicitly at the time of connection in order to gain access.

stroilova:
This looks good! Might want to confirm AWS S3 language with @mulby  or @shnayder 
:+1:

en_us/data/source/internal_data_formats/credentials.rst
```diff
@@ -221,9 +221,18 @@ Browser. Alternatively, you can use the `AWS Command Line Interface`_.
    that account. For more information, refer to the documentation for the tool
    that you selected.
 
-   Data package files are in the edX **course-data** and
-   **edx-course-data** buckets. For information about the files that you
-   download from Amazon S3, see :ref:`Package`.
+#. Select the **s3://course-data** bucket to access the database data files. 
+   
+   Select the **s3://edx-course-data/{org}** bucket to access the event data
```
mulby: This is fairly minor, but I figured I would explain it all anyway. Feel free to leave this as is if you want.

I don't think this URL should be called a "bucket". In AWS s3 lingo, only the "edx-course-data" string refers to the bucket. Buckets can contain keys. The keys are not stored hierarchically (like a filesystem) instead they have names like "foo/bar/baz.txt" which indicates to tools that all of the keys that start with "foo/" should be organized into a virtual folder of sorts called "foo". Note that no actual folder exists, the tools just do this grouping for you as a convenience. Different tools handle this differently, so depending on what tool you are using to view the S3 bucket, you may have a different experience.

I would recommend calling the "{org}" segment of the path the "folder" instead of the "bucket", since "bucket" very explicitly refers to only the "edx-course-data" part of the URL.

mulby:
perfect, thanks! :+1:

stroilova:
Yes, this is great! Thank you @mulby  and @lamagnifica  :+1: times 2.

PR #176, opened 2015-02-24
Birch RN fix for RTD

mhoeber:
remove reusables ignore path


mhoeber:
@lamagnifica or @singingwolfboy   fixing space in doc path that was breaking birch release notes on RTD.

singingwolfboy:
:+1:

PR #181, opened 2015-02-26
Fix for DOC-1276

catong:


catong:
@sstack22 thanks for reviewing, and welcome to github!

PR #185, opened 2015-03-04
diff-cover moved

nedbat:
@mhoeber not sure you want approval for stuff like this, but here you go :)


mhoeber:
:+1:   I will fix build error in other PR

PR #186, opened 2015-03-04
Sylvia/relnotes 3 4

srpearce:
@mhoeber @lamagnifica @catong Please review.

### DIFF
--- a/en_us/release_notes/source/analytics/index.rst
+++ b/en_us/release_notes/source/analytics/index.rst
@@ -3,6 +3,19 @@ edX Analytics
 ####################################
 
 ******************
+March 5, 2015
+******************
+
+* Data packages now include the following events. For more information, see
+  the `Tracking Logs`_ chapter in the `edX Research Guide`_.
+
+  * Events that record when learners view the contents of a Google Calendar or
+    Google Drive component in the LMS.
+
+  * Events that record new posts, responses, and comments in course
+    discussions.
+
+******************
 February 23, 2015
 ******************
 
--- a/en_us/release_notes/source/coming_soon.rst
+++ b/en_us/release_notes/source/coming_soon.rst
@@ -14,8 +14,6 @@ released, we will add a link to related documentation.
 * Open response assessments (ORA) in edx.org and edge.edx.org courses will soon
   support multiple prompts.
 
-* Course teams will soon be able to use new component types to include Google
-  Calendars and Google Drive documents in courseware.
 
 
 .. include:: links.rst
--- a/en_us/release_notes/source/lms/index.rst
+++ b/en_us/release_notes/source/lms/index.rst
@@ -3,6 +3,16 @@ edX Learning Management System
 ####################################
 
 ******************
+March 5, 2015
+******************
+
+* Course teams can now include Google Drive files and calendars in courseware.
+  For more information, see `Google Docs and Calendars_Students`_.
+
+* Several accessibility improvements have made navigating course content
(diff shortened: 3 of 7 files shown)


en_us/release_notes/source/coming_soon.rst
```diff
@@ -6,6 +6,9 @@ The following list describes features that are currently in development and
 that edX expects to release in the near future. After these features are
 released, we will add a link to related documentation.
 
+* Learners in self-paced courses will soon be able to generate their own
+  certificates on demand.
```
srpearce: @griffresch, do we want to announce this yet?

en_us/release_notes/source/lms/index.rst
```diff
@@ -3,6 +3,15 @@ edX Learning Management System
 ####################################
 
 ******************
+March 4, 2015
+******************
+
+* Course teams can now include Google Drive files and calendars in courseware.
```
lamagnifica, line 9: add an xref to more info?

en_us/release_notes/source/studio/index.rst
```diff
@@ -2,6 +2,11 @@
 edX Studio
 ####################################
 
+******************
+March 4, 2015
+******************
+
+* Course teams can now include Google Drive files and calendars in courseware by creating a link to the file or calendar in Studio.
```
lamagnifica: ditto on the xref suggestion. I have a strong preference to not use bulleted list formatting for lists that only have one member (I suggest using paragraph format here instead)

en_us/release_notes/source/analytics/index.rst
```diff
@@ -3,6 +3,19 @@ edX Analytics
...
+March 4, 2015
+******************
+
+* Data packages now include the following events. For more information, see
+  the Tracking Logs chapter in the `Reference Guide`_.
+
+  * Events that record when learners view the contents of a Google Calendar or
+    Google Drive component in the LMS.
+
+  * Events that record new posts, responses, and comments in the course
```
catong: I would strike "the" since you're not referring to a specific instance of some previously mentioned course discussions, but generic course discussions.

catong:
I've finished my review pass. Just one minor suggestion and agreement re adding xrefs to a couple of the notices. :+1:

lamagnifica:
:+1:

srpearce:
@explorerleslie @griffresch Please review.

explorerleslie:
@marcotuts for google drive integration. @srpearce did we clarify the doc for that to say that google isn't available in china, so be aware when adding to mooc courses?

srpearce:
@explorerleslie I've confirmed that the student-facing docs have a note about that, and I know that at one time I wrote a similar note for course authors, but I can't find it now. Aargh! I'll add that to the Google Drive files docs for course authors.

srpearce:
@explorerleslie @marcotuts Here's the note I'm adding:

Google services are not available in some regions and countries. If Google services are not available in a learner's area, the learner might see an “image unavailable” message in the place of the Google Drive file or calendar. We strongly suggest that you might provide alternative resources for learners in these areas.

explorerleslie:
:+1:

catong:
@srpearce Quick comment on your note re Google services. In the last sentence, "might" seems unnecessary. How about: "We strongly suggest that you provide alternative resources for learners in these areas."

srpearce:
@catong Good catch! I re-worded an earlier version and left "might" in.

PR #192, opened 2015-03-05
OLX Guide warning fixes, template change, change log update

mhoeber:
@lamagnifica   this gets rid of most OLX build warnings -- the ones that remain are cause of links in shared files to doc in B&R guide.  Will need longer term solution to this.

Also applied edx theme


lamagnifica:
:+1:

PR #200, opened 2015-03-10
Fix build warnings in Course Authors and Developers Guides

mhoeber:
@lamagnifica   this gets clean builds for CA and Developers Guides


lamagnifica:
:+1: thank you so much!

PR #214, opened 2015-03-14
Fixes DOC-553 include glossary at end of all guides

lamagnifica:
@mhoeber, I made the links work across guides (only worked within B&R
before). Added to Research, Developers, and OLX.


mhoeber:
:+1: looks great thank you

PR #215, opened 2015-03-15
Added reorganized front matter section, fixes DOC-1702

lamagnifica:
also added “numbered” directive to TOC trees where it was missing


lamagnifica:
@mhoeber I built the html for every affected file without error, you mat want to do the same to test.

mhoeber:
@lamagnifica  looks good, thanks.  I updated string in CA guide

PR #216, opened 2015-03-16
Insights using EMR on AWS

mulby:
I don't know why I wrote this, but I did... figured I would at least let you all know about it so as not to duplicate efforts. Feel free to ignore for the time being.

It is also incomplete, this is really only the first step (setting up all the AWS stuff). It also doesn't go into much detail about how to configure things, just tells the user what they need to do at a high level. Some of these steps are complicated...

@lamagnifica

### DIFF
--- a/en_us/install_operations/source/analytics/elastic_mapreduce.rst
+++ b/en_us/install_operations/source/analytics/elastic_mapreduce.rst
@@ -1,0 +1,363 @@
+
+.. _Using Elastic MapReduce on AWS:
+
+******************************************
+Using Elastic MapReduce on AWS
+******************************************
+
+This topic provides an overview of the components and services that you set up
+and configure in a deployment that uses Amazon EMR. Work to prepare complete
+installation procedures for edX Insights is in progress.
+
+.. contents::
+ :local:
+ :depth: 1
+
+For more information about AWS, including detailed procedures, see the `AWS
+Documentation`_.
+
+.. important::
+ The tasks described by this topic rely on the use of third-party services and
+ software. Because the services and software are subject to change by their
+ owners, the information provided here is intended as a guideline and not as
+ exact procedures.
+
+======================
+Virtual Private Cloud
+======================
+
+Create a Virtual Private Cloud (VPC) with at least one public subnet. A
+limitation of EMR running in a VPC is that clusters must be deployed into
+public subnets of VPCs. An existing VPC can be used if one is already available
+for use.
+
+EdX recommends that you configure the VPC to have at least one private subnet
+in addition to the required public subnet. A private subnet is not required.
+For more information, see the `example configuration scenario`_
+in the *Amazon Virtual Private Cloud User Guide* published by AWS.
+
+An example configuration that includes only `a single public subnet`_
+can also be found in the *Amazon Virtual Private Cloud User Guide*.
+
+Other Considerations
+*********************
+
+* To take advantage of price fluctuations in the `spot pricing market`_ , you
+  can also deploy several public subnets in different availability zones.
+
+* Consider that the clusters deployed using EMR will need network connectivity
+  to a read replica of the LMS database. EdX recommends that you use the same
+  VPC as the LMS if possible.
+
+* While it is possible to run all of these services outside of a VPC, edX does
+  not recommend doing so.
+
+===============================
+Identity and Access Management
+===============================
+
+For Amazon Identity and Access Management (IAM), create an IAM role for use
+by the EMR cluster. Assign all of the Elastic Compute Cloud (EC2) nodes in
+the cluster to this role. An option is to consider copying the contents of the
+`default EC2 role for Amazon EMR`_ used by AWS.
+
+Make sure that an IAM user with administrative privileges is available for use.
+
+=================
+SSH Credentials
+=================
+
+Generate a secure SSH key that you use to access all AWS resources. For
+example, run the following command.
+
+    .. code-block:: bash
+
+        ssh-keygen -t rsa -C "your_email@example.com"
+
+Upload the public key from the SSH key pair to AWS and assign a key pair
+name to it.
+
+======================
+Elastic Compute Cloud
+======================
+
+Ensure that at least one Amazon Elastic Compute Cloud (EC2) instance is
+available to host the various services. Depending on the scale of your
+deployment, additional instances might be necessary.
+
+Make sure to use the secure key pair name to deploy all of the EC2 instances.
+
+==============================
+Relational Database Service
+==============================
+
+Deploy a MySQL 5.6 Relational Database Service (RDS) instance. This RDS
+instance is the result store.
+
+Connectivity
+***************
+
+Ensure that there is connectivity between the EC2 instance that hosts
+the edX Analytics API and your RDS instance.
+
+Also, ensure that instances deployed into the public subnet of the VPC
+can connect to your RDS instance.
+
+Users
+***********
+
+Create at least the following two users on the RDS instance.
+
+* The "pipeline" user must have permission to create databases, tables, and
+  indexes, and issue arbitrary Data Manipulation Language (DML) commands.
+
+* The "api" user must have read-only access.
+
+Configure passwords for both of these users.
+
+Other Considerations
+*********************
+
+Configure the RDS instance to use "utf8_bin" collation by default for
+columns in all databases and tables.
+
+==================================
+Scheduler Service
+==================================
+
+Establish an SSH connection to the EC2 instance within the VPC that will
+run the scheduler service. Then, issue the following commands from a shell
+running on that instance.
+
+#. Check out the sources files from `edx-analytics-configuration`_.
+
+    .. code-block:: bash
+
+        git clone https://github.com/edx/edx-analytics-configuration.git
+
+#. Configure the shell to use the AWS credentials of the administrative AWS
+   user.
+
+    .. code-block:: bash
+
+        export AWS_ACCESS_KEY_ID=<access key ID goes here>
+        export AWS_SECRET_ACCESS_KEY=<secret access key goes here>
+
+#. Install the `AWS Command Line Interface`_.
+
+    .. code-block:: bash
+
+        pip install awscli
+
+==================================
+Simple Storage Solution Buckets
+==================================
+
+Create a Simple Storage Solution (S3) bucket to hold all Hadoop logs from
+the EMR cluster.
+
+    .. code-block:: bash
+
+        aws s3 mb s3://<your logging bucket name here>
+
+Then, create an S3 bucket to hold secure configuration files and
+initialization scripts.
+
+    .. code-block:: bash
+
+        aws s3 mb s3://<your configuration bucket name here>
+
+=========================
+MySQL Connector Library
+=========================
+
+Download the `MySQL connector library`_ from Oracle. After the download is
+complete, you then upload it to S3.
+
+  .. code-block:: bash
+
+    aws s3 cp /tmp/mysql-connector-java-5.1.*.tar.gz s3://<your configuration bucket name here>/
+
+The ``edx-analytics-configuration/batch/bootstrap/install-sqoop`` script
+references a specific version of the MySQL connector library. Update this
+``install-sqoop`` script to point to the correct version of the library in the
+S3 bucket. You must update the script before you continue.
+
+Then, upload the contents of the ``edx-analytics-configuration/batch/bootstrap/`` directory into your configuration bucket.
+
+  .. code-block:: bash
+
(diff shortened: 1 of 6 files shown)


en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -224,3 +224,275 @@ remainder of the infrastructure. The edX Analytics API and edX Insights services
 are each deployed to at least one server. The Scheduler is deployed to another
 server. A MySQL database is deployed to a server that is configured to host a
 relational database.
+
+
+******************************
+Using Elastic MapReduce on AWS
+******************************
+
```
lamagnifica: I'd suggest making this a prerequisite rather than the first step. Will this work?

=(subheading marks)==
Set Up a VPC with a Public Subnet 
=(subheading marks)==

In order to use Elastic MapReduce (EMR) on AWS, you must have a Virtual Private Cloud (VPC) with at least one public subnet. A limitation of EMR running in a VPC... If you already have a VPC with a public subnet, you can use it.
(then the bullets)

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -224,3 +224,275 @@ remainder of the infrastructure. The edX Analytics API and edX Insights services
...
+    #) It can also be advantageous to deploy several public subnets in different
+       availability zones in order to take advantage of price fluctuations in
+       the `spot pricing market <http://aws.amazon.com/ec2/purchasing-options
+       /spot-instances/>`_.
+    #) Consider that the clusters deployed using EMR will need network
+       connectivity to a read replica of the LMS database. Using the same VPC as
+       the LMS is recommended if possible.
+    #) It is possible to run all of these services outside of a VPC but not
+       recommended.
+
```
lamagnifica: going forward, can these be cast more actively, as steps to complete?
#. Optionally, copy the contents of the default IAM role. The EMR cluster creates this role and assigns all EC2 nodes to it automatically. If you copy the default IAM role, you can ... (what does this enable them to do/why would they want to copy it? change some setting, reassign nodes?)
#. Verify that an IAM user ...
#. Identify (? Locate? Save?) the secure ssh key. You use this key to access...

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -224,3 +224,275 @@ remainder of the infrastructure. The edX Analytics API and edX Insights services
...
+#. An IAM user with administrative privileges is available for use.
+
+#. A secure ssh key is generated. This will be used to access all AWS resources.
+
+    .. code-block:: bash
+
+        ssh-keygen -t rsa -C "your_email@example.com"
+
+#. The public key from the SSH key pair is uploaded to AWS and assigned a
```
lamagnifica: this doesn't seem like a step in itself, but the result of, or part of, the previous step. Indent instead of numbering?

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -224,3 +224,275 @@ remainder of the infrastructure. The edX Analytics API and edX Insights services
...
+   keypair name.
+
+#. Ensure at least one EC2 instance is available to host the various services.
+   Depending on the scale of the deployment, more instances may be necessary.
```
lamagnifica: may --> might

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -224,3 +224,275 @@ remainder of the infrastructure. The edX Analytics API and edX Insights services
...
+   The instances should be deployed using the secure keypair name.
```
lamagnifica: I'd avoid phrasing with "should" or "may" and instead tell them the right thing to do... 
"Be sure to use the secure keypair name when you deploy any additional instances."

en_us/install_operations/source/analytics/install_analytics.rst
lamagnifica: ok, I think I understand your purpose for this section better. Rather than a procedure for doing setup, what you are offering in this new section is a checklist to make sure that an environment is set up appropriately. I think a different title would help get that across better: Preparing to Use... or something like that?

lamagnifica:
@mulby, I think this is a great start, and I'm sure if it will offload some of the support questions you've been fielding. My main observation is to more clearly define what task the readers are setting out to complete, or what problem they're trying to solve, that would bring them to this section. That will help us title it so that they can find it, and also help us decide if this is a procedure with ordered steps to follow, or a collection of prerequisites or requirements. We may want to explicitly point people back up to the list of "what you should know" before we start describing either one. 
I'd be glad to work at changing the phrasing from passive to active voice if that will offload some effort from you. Let me know how I can help.

mhoeber:
@mulby  @lamagnifica   going through older PRs.  Possible we can get this done for Cypress release?

mulby:
@mhoeber I would love to put more time into this, but given that it's not prioritized by product I push this type of thing forward in my spare time. I'd be OK with you closing this for now, until it is prioritized.

Note that the analytics stack (for better or worse) is not included in the edX named releases (at least for the time being). We should revisit that decision after Cypress, but for now I'm not sure there is much value in pushing this in before Cypress is finished.

@shnayder FYI

mhoeber:
@mulby   yes, makes sense.  I'll put something in our backlog to look at for the "D" release.

lamagnifica:
@mulby @shnayder I copy edited this, with the interesting stuff starting at line 233. I added headings  to provide more organizational cues, which are definitely just guess work as a starting point. I also changed it to active voice, present tense, and the indicative mood (mostly). I didn't intentionally change the meaning of any of Gabe's original draft! Let me know what else I can do to get this ready for publication.

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -224,5 +230,357 @@ services are each deployed to at least one server. The Scheduler is deployed to
...
+You must ensure that there is connectivity between the EC2 instance that hosts
+the edX Analytics API and your RDS instance.
+
+You must also ensure that instances deployed into the public subnet of the VPC
+can connect to your RDS instance.
+
+Users
+***********
+
+You create at least these users on the RDS instance.
```
lamagnifica: these --> the following two

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -224,5 +230,357 @@ services are each deployed to at least one server. The Scheduler is deployed to
...
+    .. code-block:: bash
+
+        aws s3 cp /tmp/mysql-connector-java-5.1.*.tar.gz s3://<your configuration bucket name here>/
+
+The ``edx-analytics-configuration/batch/bootstrap/install-sqoop`` script
+references a specific version of the MySQL connector library. You update this
+``install-sqoop`` script to point to the correct version of the library in the
+S3 bucket. You must update the script before you continue.
+
+Finally, you upload the contents of the
```
lamagnifica: Finally, --> Then,

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -92,22 +94,23 @@ access the data using a supported web browser. EdX Insights communicates
...
-* Understand basic terminal usage.
-* Understand how the LMS has been deployed and configured.
-* Understand basic computer network terminology.
-* Understand the YAML file format.
-* Understand Amazon Web Services terminology.
+* Basic terminal usage.
+* How the LMS has been deployed and configured.
+* Basic computer network terminology.
+* YAML file format.
+* Amazon Web Services (AWS) terminology.
```
shnayder: (If you plan to use AWS services)

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -124,21 +127,22 @@ Hadoop
 
 Most of the computation performed by the edX Analytics Pipeline is implemented
 as Map Reduce jobs that currently must be executed by a Hadoop cluster. You can
```
shnayder: remove "currently"

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -162,7 +166,7 @@ The results of computations performed by the edX Analytics Pipeline are stored
 in a MySQL database. Even small installations should use a different MySQL
 server than the one used by the LMS. The query patterns of the edX Analytics
```
shnayder: it's really the pipeline's write patterns to the result store that can cause problems, not the API's query patterns. (Right @mulby ?)

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -224,5 +230,357 @@ services are each deployed to at least one server. The Scheduler is deployed to
...
+available to host the various services. Depending on the scale of your
+deployment, additional instances might be necessary.
+
+Be sure to use the secure key pair name to deploy all of the EC2 instances.
+
+==============================
+Relational Database Service
+==============================
+
+You deploy a MySQL 5.6 Relational Database Service (RDS) instance. This RDS
```
shnayder: How about just "Deploy...", and similarly for the other "You <verb>" instances?

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -224,5 +230,357 @@ services are each deployed to at least one server. The Scheduler is deployed to
...
+        skipping: [10.0.1.236]
+
+        TASK: [user | create allowed command links] ***********************************
+        skipping: [10.0.1.236]
+
+        PLAY RECAP ********************************************************************
+        10.0.1.236                 : ok=0    changed=4    unreachable=0    failed=0
+        localhost                  : ok=4    changed=1    unreachable=0    failed=0
+
+
```
shnayder: We should add a note that there's more to do after this -- setting up the scheduler, configuring all the things, etc.

Perhaps link to the wiki page from the hackathon as a resource for those steps for now?

shnayder:
I took a pass. I think we could add a bit of text to the top to make it clear that this is a work in progress, and then merge it.  @mulby, what do you think?

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -224,5 +230,357 @@ services are each deployed to at least one server. The Scheduler is deployed to
...
+        aws s3 cp /tmp/mysql-connector-java-5.1.*.tar.gz s3://<your configuration bucket name here>/
+
+The ``edx-analytics-configuration/batch/bootstrap/install-sqoop`` script
+references a specific version of the MySQL connector library. You update this
+``install-sqoop`` script to point to the correct version of the library in the
+S3 bucket. You must update the script before you continue.
+
+Finally, you upload the contents of the
+``edx-analytics-configuration/batch/bootstrap/`` directory into the
+``openedx-analytics-configuration`` bucket.
```
brianhw: This is a different name for the bucket.  Call it "your configuration bucket"?

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -224,5 +230,357 @@ services are each deployed to at least one server. The Scheduler is deployed to
...
+                master: {
+                    num_instances: 1,
+                    type: m1.medium,
+                    market: ON_DEMAND,
+                },
+                core: {
+                    num_instances: 2,
+                    type: m1.medium,
+                    market: SPOT,
+                    bidprice: 0.03
```
brianhw: Do you want to highlight that the bidprice and the type may be values that they want to change?

en_us/install_operations/source/analytics/install_analytics.rst
```diff
@@ -224,5 +230,357 @@ services are each deployed to at least one server. The Scheduler is deployed to
...
+                {
+                    type: script,
+                    name: Install Sqoop,
+                    step_args: [
+                        "s3://<your configuration bucket name here>/install-sqoop",
+                        "s3://<your configuration bucket name here>"
+                    ]
+                }
+            ],
+            user_info: []
```
brianhw: will this work without user_info?

lamagnifica: @mulby ^^ ?

en_us/install_operations/source/analytics/elastic_mapreduce.rst
```diff
@@ -0,0 +1,363 @@
...
+                    type: script,
+                    name: Install Sqoop,
+                    step_args: [
+                        "s3://<your configuration bucket name here>/install-sqoop",
+                        "s3://<your configuration bucket name here>"
+                    ]
+                }
+            ],
+            user_info: []
+        }
```
lamagnifica, line 254: @mulby , @brianhw asks, "will this work without user_info?"

PR #217, opened 2015-03-16
DOC: Glossary updates

catong:
@lamagnifica can you please review? I added "Content Libraries", "Libraries". Also updated "Import" to include mention of libraries, and added "Export".

### DIFF
--- a/en_us/shared/glossary/glossary.rst
+++ b/en_us/shared/glossary/glossary.rst
@@ -105,8 +105,8 @@ C
   discussions.
 
   Cohorts are an optional feature of courses on the edX platform. For
-  information about how you enable the cohort feature, set up cohort groups,
-  and assign students to them, see `Using Cohorts in Your Courses`_.
+  information about how you enable the cohort feature, set up cohorts, and
+  assign students to them, see `Using Cohorts in Your Courses`_.
 
 .. _Component_g:
  
@@ -125,6 +125,12 @@ C
   use content experiments to compare the performance of students who have been
   exposed to different versions of the content. For more information, see
   `Creating Content Experiments`_.
+
+
+**Content Library**
+
+  See :ref:`Library<L>`.
+
 
 **Content-Specific Discussion Topic**
 
@@ -295,6 +301,17 @@ E
   the learner engaged. Exercises are also an important measure of teaching
   effectiveness and learner comprehension.
 
+
+.. _Export:
+ 
+**Export**
+
+  A tool in edX Studio that you use to export your course or library for
+  backup purposes, or so that you can edit the course or library directly in
+  XML format. See also :ref:`Import<I>`.
+
+  For more information, see `Export a Course`_ or `Export a Library`_.
+
 .. _F:
 
 ***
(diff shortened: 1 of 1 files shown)


en_us/shared/glossary/glossary.rst
```diff
@@ -105,8 +105,8 @@ C
   discussions.
 
   Cohorts are an optional feature of courses on the edX platform. For
-  information about how you enable the cohort feature, set up cohort groups,
-  and assign students to them, see `Using Cohorts in Your Courses`_.
+  information about how you enable the cohort feature, set up cohorts, and
+  assign students to them, see `Using Cohorts in Your Courses`_.
```
lamagnifica, line 109: oh, good catch

en_us/shared/glossary/glossary.rst
```diff
@@ -295,6 +301,16 @@ E
   the learner engaged. Exercises are also an important measure of teaching
   effectiveness and learner comprehension.
 
+
+.. _Export:
+ 
+**Export**
+
+  A tool in edX Studio that allows you to export your course or library for backup purposes, or so that you can edit the course or library directly in
```
lamagnifica: tool --> feature?
wrapping
this is so great to add

en_us/shared/glossary/glossary.rst
```diff
@@ -359,15 +375,17 @@ I
...
 
+
 .. _Import:
  
 **Import**
 
-  A tool in edX Studio that loads a new course into your existing course. When
-  you use the Import tool, Studio replaces all of your existing course content
-  with the content from the imported course.
+  A tool in edX Studio that loads a new course or library into your existing
```
lamagnifica: feature?

en_us/shared/glossary/glossary.rst
catong: I looked at "tool" (in the original) and thought I'd leave it. It seemed more user-centric in this case than "feature", which is more abstract and can be seen as coming from a product marketing/dev point of view. Some groups of functionality around a concept (e.g. cohorts) can be called features for lack of a better term, but for something as concrete as import and export, I think "tool" is fine.
I'll fix line wrap.

en_us/shared/glossary/glossary.rst
```diff
@@ -359,15 +375,17 @@ I
...
+  course or library. When you use the Import tool, Studio replaces all of your
```
lamagnifica: ? into your existing course or library?
I think the original definition was a bit off. Maybe,
Import is a feature that allows you to add a course or library written in XML format, such as a previously exported course or library, to edX Studio. If you import a course or library that has the same identifying information (I'm not sure what this is, probably name, org, number, run) as an existing course, the imported course replaces all of the existing course content.

en_us/shared/glossary/glossary.rst
```diff
@@ -404,6 +422,17 @@ L
   and access data while the course is running.
 
 
+.. _Library:
+
+**Library**
+
+  A pool of components for use in randomized assignments that can be shared
```
lamagnifica, line 430: I would remove the "for use in randomized assignments" clause

en_us/shared/glossary/glossary.rst
```diff
@@ -404,6 +422,17 @@ L
...
+  across multiple courses from your organization. Course teams configure
+  randomized content blocks in course outlines to randomly provide a specified
+  number of problems to each student.
```
lamagnifica: 2nd sentence isn't explicitly connected to the first sentence. Maybe, 
Course teams configure randomized content block components to reference a library. When a randomized content block is included in the outline of a course, it randomly provides a specified number of problems from the library to each student.

lamagnifica:
:+1: thanks for adding extra definitions!

en_us/shared/glossary/glossary.rst
catong, line 430: Per Product, this is the use case we want to clearly emphasize, because currently you can't use libraries in any other way.

catong:
Merging https://github.com/edx/edx-documentation/pull/217

en_us/shared/glossary/glossary.rst
lamagnifica, line 430: my thinking was to keep the description a bit more general here, and let the pointer to the rest of the doc provide the detail. As soon as any other library-related feature is released, this description will have to be updated...

PR #219, opened 2015-03-16
CA Change Log March 17 release items

catong:
Adding change log updates for items included in March 17 release


catong:
Merging https://github.com/edx/edx-documentation/pull/219

PR #225, opened 2015-03-18
Update CA change log for Mar 17

catong:
Update change log for course authors to include Content Libraries and Randomized Content Blocks


catong:
Merging https://github.com/edx/edx-documentation/pull/225

PR #236, opened 2015-03-23
Adds new "Video Available on Web Only" setting

lamagnifica:
see lines 321-324.
@gwprice , @catong , @lou-wang , @jhendersonedx 
Fixes DOC-1765

### DIFF
--- a/en_us/shared/building_and_running_chapters/creating_content/create_video.rst
+++ b/en_us/shared/building_and_running_chapters/creating_content/create_video.rst
@@ -8,9 +8,9 @@ Working with Video Components
 Introduction to Videos
 **********************
 
-You can create a video of your lecture and add it to your course with other
-components—such as discussions and problems—to promote active learning. Adding
-a video to your course has several steps.
+You can create videos of your lectures and add them to your course to
+supplement other components, such as discussions and problems, to promote
+active learning. Adding a video to your course has several steps.
 
 #. :ref:`Create the Video`.
 #. :ref:`Create Transcript`.
@@ -150,9 +150,9 @@ accesses your YouTube videos.
 
 Because YouTube is not available in all locations, however, we recommend that
 you also post copies of your videos on a third-party site such as `Amazon S3
-<http://aws.amazon.com/s3/>`_. When a student views  a video in your course, if
-YouTube is not available in that student’s location or if the YouTube video
-doesn’t play, the video on the backup site starts playing automatically. You
+<http://aws.amazon.com/s3/>`_. When a student views a video in your course, if
+YouTube is not available in that student's location or if the YouTube video
+does not play, the video on the backup site starts playing automatically. You
 can also allow the student to download the video from the backup site.
 
 After you post your video online, make sure you have the URL for the video. If
@@ -200,12 +200,12 @@ Step 4. Create a Video Component
 
    You replace the default values with your own. 
    
-3. In the **Component Display Name** field, enter the name you want students to
-   see when they hover the mouse over the unit in the course ribbon. This text
-   also appears as a header for the video.
+3. In the **Component Display Name** field, enter an identifying name for this
+   video. Students see this name as a header for the video and when they hover
+   the cursor over the unit in the course ribbon.
 
-#. In the **Default Video URL** field, enter the URL of the video. For example,
-   the URL may resemble one of the following.
+#. In the **Default Video URL** field, enter the URL of the video. Example
+   URLs follow.
 
    ::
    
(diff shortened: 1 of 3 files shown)


en_us/shared/building_and_running_chapters/creating_content/create_video.rst
```diff
@@ -9,8 +9,8 @@ Introduction to Videos
 **********************
 
 You can create a video of your lecture and add it to your course with other
-components—such as discussions and problems—to promote active learning. Adding
-a video to your course has several steps.
+components, such as discussions and problems, to promote active learning.
```
catong: The wording of "..add it to your course with other components" is hard to parse and unnecessarily confusing  -- it's as if we are talking about adding the other types of components at the same time as the video. If we want to mention the other components I'd recommend rewording... talk about adding video to _supplement_ other components in your course.

en_us/shared/building_and_running_chapters/creating_content/create_video.rst
```diff
@@ -200,12 +200,12 @@ Step 4. Create a Video Component
 
    You replace the default values with your own. 
    
-3. In the **Component Display Name** field, enter the name you want students to
-   see when they hover the mouse over the unit in the course ribbon. This text
-   also appears as a header for the video.
+3. In the **Component Display Name** field, enter an identifying name for this
```
catong, line 203: How about "enter a name to identify this video."

en_us/shared/building_and_running_chapters/creating_content/create_video.rst
```diff
@@ -220,13 +220,13 @@ Step 4. Create a Video Component
...
+    student's computer plays.
 
 5. Next to **Default Timed Transcript**, select an option. 
 
-   * If edX already has a transcript for your video--for example, if you are
-     using a video from an existing course--Studio automatically finds the
-     transcript and associates the transcript with the video.
+   * If edX already has a transcript for this video, Studio automatically
+     finds the transcript and associates the transcript with the video. This
+     typically occurs when you are reusing a video from an existing course.
```
catong: "This ^situation^ typically occurs..."

en_us/shared/building_and_running_chapters/creating_content/create_video.rst
```diff
@@ -318,6 +318,10 @@ The following options appear on the **Advanced** tab in the video component.
       - Allows you to upload a handout to accompany this video. Your handout
         can be in any format. Students can download the handout by clicking
         **Download Handout** under the video.
+    * - **Video Available on Web Only**
+      - If True, indicates that students are only allowed to play this video
```
catong: Do we say "If True" everywhere else? Typically I prefer "If the value of this setting is True"... more verbose but better for translation.

catong: Not sure if there is a nuance to "are only allowed to play". How about "can only" instead, if there's really no permission-related issue.

en_us/shared/building_and_running_chapters/creating_content/create_video.rst
```diff
@@ -318,6 +318,10 @@ The following options appear on the **Advanced** tab in the video component.
...
+        in a Web browser. If False, students can use any compatible
```
catong: Same as comment above: "If the value of this setting is False" (if not inconsistent with other occurrences in this file)

catong:
@lamagnifica Looks good; :+1: with some suggestions.

en_us/shared/building_and_running_chapters/creating_content/create_video.rst
lamagnifica, line 203: the "enter an identifying name" phrasing was recently added to all of the files in the developing_course directory. I'd rather stick with the same phrasing for components as well, although it is still too soon to say that the name of a non-problem will be surfaced in Insights...

en_us/shared/building_and_running_chapters/creating_content/create_video.rst
lamagnifica: It is a legal requirement, often because a video contains material that is being used by permission, and that permission only grants them the right to distribute through the edX web site. But, since this is for open source too, we can't use that as an example...

en_us/shared/building_and_running_chapters/creating_content/create_video.rst
lamagnifica: This file does not currently tell you what happens if you set any of the other  advanced settings to true or to false. I would like to move in that direction, as I think it alleviates the, well, which one do I set it to then? response to a description. It is not my intention to research and rewrite all of these settings for this PR however....

en_us/shared/building_and_running_chapters/creating_content/create_video.rst
lamagnifica: I think the phrasing, "If the value of this setting is True..." assumes that you are reviewing an existing setting for reference, rather than actively selecting a setting. As an alternative to the "If True/If False" construction, maybe "If {whatever}, select True."?

catong: I like it! "If (you) select True..."

gwprice:
:+1:

PR #242, opened 2015-03-24
Fixes DOC-1784 and AN-3921

lamagnifica:
@shnayder please review.


lamagnifica:
@stroilova if you're interested

en_us/data/source/internal_data_formats/sql_schema.rst
```diff
@@ -568,6 +568,8 @@ mailing_address
   Collected during student registration from a text field control. A blank
   string for students who elect not to enter anything.
 
+  This column can contain multiple lines, which are separated by '\r'.
```
stroilova: Hi @lamagnifica  , I checked this this in vim, emacs, and cat on Mac, and it appears as '^M\n' which I think is CR+LF , or '\r\n'

http://en.wikipedia.org/wiki/Newline

FYI @mulby  @shnayder in case I'm saying this wrong.

PS>> Alison, I am not sure if the backslash is rendering properly in the HTML view -- perhaps it should be monospaced?

en_us/data/source/internal_data_formats/sql_schema.rst
```diff
@@ -628,6 +630,8 @@ goals
   label "Goals in signing up for edX". A blank string for students who elect
   not to enter anything.
 
+  This column can contain multiple lines, which are separated by '\r'.
```
stroilova: '\r\n'

en_us/data/source/internal_data_formats/sql_schema.rst
```diff
@@ -568,6 +568,8 @@ mailing_address
   Collected during student registration from a text field control. A blank
   string for students who elect not to enter anything.
 
+  This column can contain multiple lines, which are separated by '``\r\n``'.
+
```
lamagnifica, line 572: @stroilova please let me know if this format works for you, or if you prefer the apostrophe characters to be mono as well

stroilova, line 572: Looks good. We can change it later if people ask more questions, but this makes sense.  :+1:

PR #244, opened 2015-03-25
Remove extra slash in URL for randomized content blocks

srpearce:
@mhoeber @lamagnifica @catong FYI, this fixes the bug Melanie reported (extra slash in the Randomized Content Blocks URL) - no review needed


catong:
Thanks for taking care of this!

PR #271, opened 2015-04-03
Add new projects to testing

mhoeber:
@clytwynec   added new projects to testing.  Is this all that's required to have Travis build these new projects as well?


clytwynec:
Yup.  :+1:  It looks like it is failing because theres an error in one of the sources, but these changes are working fine.

PR #277, opened 2015-04-06
Fixes DOC-1623

lamagnifica:
@catong would you give this a look? ECOM-579 corrects the label on the
Drupal site


catong:
Yay! :-) :+1:

PR #281, opened 2015-04-07
Fixes DOC-1788

lamagnifica:
@mhoeber @lou-wang @gwprice please take a look.

### DIFF
--- a/en_us/video/source/video_uploads.rst
+++ b/en_us/video/source/video_uploads.rst
@@ -206,6 +206,10 @@ The encoding and hosting process assigns these statuses to video files.
   Processing continues at video hosting sites for 24 hours after you upload a
   file.
 
+* **Imported** files were imported into Studio along with the other assets of
+  a previously exported course. Imported files are often already included
+  in course components, and they are ready for learners to view in the LMS.
+
 * **Failed** files did not complete processing successfully. Verify that you
   can play your original .mp4 or .mov file and that it meets the other
   specifications for successful video processing. See :ref:`Specifications for
--- a/en_us/video/source/video_uploads.rst
+++ b/en_us/video/source/video_uploads.rst
@@ -207,8 +207,12 @@ The encoding and hosting process assigns these statuses to video files.
   file.
 
 * **Imported** files were imported into Studio along with the other assets of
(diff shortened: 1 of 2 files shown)


en_us/video/source/video_uploads.rst
```diff
@@ -206,6 +206,14 @@ The encoding and hosting process assigns these statuses to video files.
   Processing continues at video hosting sites for 24 hours after you upload a
   file.
 
+* **Imported** files were imported into Studio along with the other assets of
+  a previously exported course. To determine whether processing is complete
+  for a file with a status of **Imported**, refer to the status of the same
+  file on the **Video Uploads page of the original course.
```
gwprice, line 212: Missing the trailing *\* on Video Uploads?

gwprice:
The reality is actually a little more complicated. Part of the reason we didn't want to import the status from the source Open edX instance is that the source and destination instances may have completely different video profiles and completely different notions of what is and is not "ready". For example, another instance may use an alternative hosting site instead of Youtube and not consider a video ready until it has been uploaded there. In that case, a video that was ready when exported from edx.org may not be considered ready when imported into another system. Unfortunately, we don't currently have a way to reconcile such discrepancies between deployments, which is why we settled on the "imported" status label. I think the best advice we can give is that any imported video should be tested on a variety of devices and connections to verify that the import is functional.

lamagnifica:
@gwprice apologies, I should have made it clear that the document that I am editing here is the one that is private, shared with our partner organizations only. We do not currently have any outward-facing documentation for the Video Uploads page, and such a document is not currently planned AFAIK. So I am hopeful that the advice given here is useful for our partner teams, using the VAL VEDA pipeline.

gwprice:
Ah. For edX users specifically, I doubt that this would actually come up in the near- to medium-term future, because it would require building a course on and uploading videos to another Open edX instance and then exporting that course and importing it on edx.org.

lamagnifica:
@lou-wang , in light of Greg's expected (lack of) use cases, ok to discard this?

lamagnifica:
per @lou-wang . "i'm ok leaving it out"

PR #284, opened 2015-04-08
Reference the Lavash torrent file, not Injera

singingwolfboy:
We might want to confirm this change with devops


mhoeber:
@feanil can you confirm ^^

feanil:
That sounds fine, we really need to do a new bread release but I'll let you know if that happens.

PR #293, opened 2015-04-14
Fix course structures link

mhoeber:
@lamagnifica   updated to fix broken links;  for an unknown reason, "Course Structures" as a label breaks sphinx.  this is a workaround


lamagnifica:
:+1:

PR #296, opened 2015-04-14
Replaces ribbon and accordion throughout

lamagnifica:
@catong @srpearce @mhoeber I made these changes to the glossary and to the text as a step toward using our new standard terms for these navigational elements.


mhoeber:
:+1:

PR #298, opened 2015-04-15
Added IE 11

lamagnifica:
@benpatterson , @mhoeber please review. 
I added IE 11 as fully supported in all three products. I kept IE9. The image that follows may be easier to read than the rst table markup.

screen shot 2015-04-15 at 2 19 20 pm

### DIFF
--- a/en_us/shared/browsers.rst
+++ b/en_us/shared/browsers.rst
@@ -33,18 +33,20 @@ learning management system (LMS), which learners and course teams use to
 interact with course content. 
 
 .. list-table::
-   :widths: 20 10 10 10 10 10
+   :widths: 20 10 10 10 10 10 10
    :header-rows: 1
 
    * -
      - Chrome
      - Safari
      - Firefox
-     - IE 9
+     - IE 11
      - IE 10
+     - IE 9
    * - Windows 8
      - Yes
      - N/A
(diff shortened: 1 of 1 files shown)


mhoeber:
:+1:

benpatterson:
Hrm, is there a reason we don't say 'provisional' like we did on the old page? @lamagnifica

lamagnifica:
@benpatterson we certainly can -- I just understood the JIRA story as saying that we did support IE11, without the provisions...!
DOC-1871

benpatterson:
Sorry for any confusion. My point was that IE11 should be on the list---following whatever pattern we have for the other IEs. In that case, we only had IEs as provisional in all cases, unless I was missing some other piece of documentation...

lamagnifica:
Revisions made:
screen shot 2015-04-15 at 3 51 21 pm

benpatterson:
:+1:

lamagnifica:
@benpatterson for comparison, this is what we currently state:
http://edx.readthedocs.org/projects/edx-release-notes/en/latest/front_matter/browsers.html
Note that we currently say that we support IE 9 and 10 for the LMS, but provisionally support them for Studio and Insights. Please let me know if that is not accurate.

PR #311, opened 2015-04-27
Reworded to make this a little clearer

lamagnifica:
@srpearce I couldn't let this go over the weekend, and have perhaps found a better place for this info and a better way to state it.


srpearce:
:+1:

PR #313, opened 2015-04-28
Learner Engagement report on inst dash

lamagnifica:
@shnayder , @brianhw , @mulby here's an initial version of the doc for the csv. Please let me know of inaccuracies, additional useful info that I've omitted, etc.
@srpearce take a look!

### DIFF
--- a/en_us/shared/building_and_running_chapters/running_course/course_student.rst
+++ b/en_us/shared/building_and_running_chapters/running_course/course_student.rst
@@ -41,7 +41,7 @@ distribution or misuse of this information.
 .. _Access_student_data:
 
 ****************************
-Access Student Data
+Accessing Student Data
 ****************************
 
 You can download data about the students who are currently enrolled in your
@@ -109,7 +109,7 @@ view student data on the Instructor Dashboard.
 .. note:: In addition to the data for enrolled students, data for the course 
  staff is included in the file or display.
 
-To download student data:
+To download student data, follow these steps.
 
 #. View the live version of your course.
 
@@ -131,7 +131,7 @@ To download student data:
 
   All student-supplied data is included in this file without truncation.
 
-To view student data:
+To view student data, follow these steps.
 
 .. note:: This option is available only for courses with an enrollment of less 
  than 200.
@@ -160,7 +160,7 @@ Cohort column with each student's assigned cohort group.
 .. _Access_anonymized:
 
 ********************************
-Access Anonymized Student IDs
+Accessing Anonymized Student IDs
 ********************************
 
 Some of the tools that are available for use with the edX platform, including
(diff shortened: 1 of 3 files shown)


srpearce:
@lamagnifica I'm on it.

en_us/shared/building_and_running_chapters/running_course/course_student.rst
```diff
@@ -188,22 +188,112 @@ the ``{course_id}_student_profile_info_{date}.csv`` file of student data or the
...
+*************************************
+
+With the learner engagement report, you can monitor what individual learners
+are doing in your course. The report contains a row for each enrolled learner,
+and has columns that quantify overall course activity and engagement with
+course problems, videos, discussions, and textbooks.
+
+With this report, you can identify which learners are, and which are not,
+visiting any page in the course. Further, you can identify the learners who
+are attempting problems, playing videos, participating in discussions, viewing
```
srpearce: Add "and" or "or" before "viewing"

en_us/shared/building_and_running_chapters/running_course/course_student.rst
```diff
@@ -188,22 +188,112 @@ the ``{course_id}_student_profile_info_{date}.csv`` file of student data or the
...
+       period.
+   * - Course ID
+     - The identifier for the course run.
+   * - Username
+     - The unique username for an edX account. 
+   * - Email
+     - The unique email address for an edX account.
+   * - Cohort
+     - For courses that have the cohorts feature enabled, this column
+       indicates the learner's assigned cohort group.
```
srpearce: Nit: I'd strike "this column indicates" (the other descriptions all start with sentence fragments, even if they then include a second complete sentence).

srpearce:
Two nits - otherwise :+1:

en_us/shared/building_and_running_chapters/running_course/course_student.rst
```diff
@@ -188,22 +188,112 @@ the ``{course_id}_student_profile_info_{date}.csv`` file of student data or the
...
+Using the Learner Engagement Report
+*************************************
+
+With the learner engagement report, you can monitor what individual learners
+are doing in your course. The report contains a row for each enrolled learner,
+and has columns that quantify overall course activity and engagement with
+course problems, videos, discussions, and textbooks.
+
+With this report, you can identify which learners are, and which are not,
+visiting any page in the course. Further, you can identify the learners who
```
brianhw: The report is not really tracking visits to pages.  It is tracking which problems are checked and which videos are viewed, but we're not actually looking at whether they visited the pages where the problem or video might be viewed.  This first sentence seems to suggest that information is provided about each page, rather than some page.  Something like "... visiting content in the course"?

en_us/shared/building_and_running_chapters/running_course/course_student.rst
```diff
@@ -188,22 +188,112 @@ the ``{course_id}_student_profile_info_{date}.csv`` file of student data or the
...
+.. note:: To measure problem-related activity, the report includes data only 
+ for problems with these problem types.
+
+ * Checkboxes (``<choiceresponse>``)
+ * Dropdown (``<optionresponse>``)
+ * Multiple choice (``<multiplechoiceresponse>``)
+ * Numerical input (``<numericalresponse>``)
+ * Text input (``<stringresponse>``)
+ * Math expression input (``<formularesponse>``)
+
```
brianhw, line 204: I'm not sure what the actual list would end up being, but the activity report is actually looking at any problem_check server event that includes a problem_id.  If the event includes "success"="correct", then it's marked as correct, otherwise it's incorrect.  So it's really only limited to CAPA problems (e.g. as opposed to ORA2 openended responses).  Does "problems" sufficiently constrain that?

Note that there are several classes of these problems for which the correctness may not be accurately reflected.  For example, any problem that submits a response to an xqueue for scoring will likely be marked as incorrect, even if it were (eventually) correct.

en_us/shared/building_and_running_chapters/running_course/course_student.rst
brianhw: Also may be helpful to indicate that the value will be a zero-length string (or blank or "" or however you say that) for learners who are not assigned to a cohort.

A small distinction:  we are outputting the user's cohort assignment regardless of whether the course has the cohort feature enabled.  That is, it is possible for the feature to be enabled, then for some users to be assigned (as they access the course), and then for the feature to be disabled.  And even if the feature were enabled, there may be many users who do not yet have an assigned cohort, and may never get one (e.g. if they fail to engage with the course again).

en_us/shared/building_and_running_chapters/running_course/course_student.rst
```diff
@@ -188,22 +188,112 @@ the ``{course_id}_student_profile_info_{date}.csv`` file of student data or the
...
+     - The unique email address for an edX account.
+   * - Cohort
+     - For courses that have the cohorts feature enabled, this column
+       indicates the learner's assigned cohort group. 
+   * - Was Active
+     - Included only in the daily report. 1 for learners who visited any page
+       (URL) in the course at least once during the reported day, 0 otherwise.
+   * - Days Active this Week
+     - Included only in the weekly report. 1 for learners who visited any page
+       (URL) in the course at least once during the reported week, 0 otherwise.
```
brianhw: "Days Active This Week" instead of "Days Active this Week".  (Did I get that wrong?)  Anyway, this is not a 0 or 1 column, but a count of the number of days for which "Was Active" would have been 1.  So maybe "+1 for learners... at least once on a given day during the reported week"?

en_us/shared/building_and_running_chapters/running_course/course_student.rst
```diff
@@ -188,22 +188,112 @@ the ``{course_id}_student_profile_info_{date}.csv`` file of student data or the
...
+   * - Unique Problems Attempted
+     - The number of unique problems for which the learner selected **Check**
+       to submit an answer. 
+   * - Total Problem Attempts
+     - The number of times the learner selected **Check**
+       to submit answers, across all problems attempted.
```
brianhw: I was a little confused by the "across all problems attempted".  Perhaps "regardless of the particular problem attempted"?

brianhw:
I've done making a pass.

en_us/shared/building_and_running_chapters/running_course/course_student.rst
lamagnifica, line 204: @shnayder , is the set of problem types that Brian describes here what you expect for this report?

en_us/shared/building_and_running_chapters/running_course/course_student.rst
lamagnifica: agreed. but I am confused by your statement @brianhw -- what about interaction with HTML component and textbooks?

en_us/shared/building_and_running_chapters/running_course/course_student.rst
```diff
@@ -188,22 +188,112 @@ the ``{course_id}_student_profile_info_{date}.csv`` file of student data or the
...
+.. _Download the Learner Engagement Report:
+
+===================================================
+Download the Learner Engagement Report
+===================================================
+
+An automated process runs daily on the system server to update learner
+engagement data and create the daily or weekly .csv file for you to download.
+A link to the most recently updated version of the .csv file is available on
+the Instructor Dashboard.
```
brianhw: Actually, links to all .csv files will be available.  They all have a separate date in the filename, and describe engagement over a separate time period.  I don't know if that's worth mentioning, but they may be surprised by the accumulation.

en_us/shared/building_and_running_chapters/running_course/course_student.rst
brianhw: You're right that maybe "content" isn't general enough.  Maybe it's enough to say "... visiting your course" and leave out the reference to "any page in".

lamagnifica:
@brianhw , @shnayder , please take another look when you get a chance.

brianhw:
LGTM. :+1:

PR #318, opened 2015-04-30
Fixes DOC-1952

lamagnifica:
@rgrubb, @mhoeber , @srpearce , @catong please take a look
https://openedx.atlassian.net/browse/DOC-1952

### DIFF
--- a/en_us/video/source/video_prereqs.rst
+++ b/en_us/video/source/video_prereqs.rst
@@ -75,10 +75,10 @@ the video administrators.
 Establishing Access to the Account 
 ===================================
 
-#. Create a single Google email, or Gmail, account. The recommended format
+#. Create a single Google email (Gmail) account. The recommended format
    for the account name is your edx.org member identifier followed by
-   "-CMSmanager". For example, ``HarvardX-CMSmanager@gmail.com`` or 
-   ``MITx-CMSmanager@gmail.com``.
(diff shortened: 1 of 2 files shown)


catong:
@mhoeber @srpearce I can look at this one.

en_us/video/source/video_prereqs.rst
```diff
@@ -77,8 +77,8 @@ Establishing Access to the Account
 
 #. Create a single Google email, or Gmail, account. The recommended format
```
catong: instead of separate clause set off with commas in this case it might be clearer to just put Gmail in parens.
"Create a single Google email (Gmail) account.

en_us/video/source/video_prereqs.rst
```diff
@@ -77,8 +77,8 @@ Establishing Access to the Account
...
    for the account name is your edx.org member identifier followed by
-   "-CMSmanager". For example, ``HarvardX-CMSmanager@gmail.com`` or 
-   ``MITx-CMSmanager@gmail.com``.
+   "CMSmanager". For example, ``HarvardXCMSmanager@gmail.com`` or 
+   ``MITxCMSmanager@gmail.com``.
```
catong, line 81: If the point behind this correction is that account names should not include hyphens, should you point this out?

catong:
Looks fine though I had a question about the reason behind changing the email account name examples. :+1:

PR #332, opened 2015-05-08
Ahodges/SD_config_mobile

lamagnifica:
I definitely recommend looking at this in split view...
@nimisha, I updated the configuration instructions to match the relevant parts of the https://openedx.atlassian.net/wiki/pages/viewpage.action?spaceKey=MA&title=Settings+for+Mobile+API+Sandbox page. I'm not sure if I've characterized the differences between configuring the json vs yml files correctly. Any help appreciated. I added push notifications to the end of the mobile.rst file -- I'd like to work on providing more detail, but only if/when you or another engineer has the time to spend, or if we need more for our June delivery.
@mhoeber , I did some editing for style guide issues, and made a suggestion about the way the contents display at the top of the mobile.rst file. Let me know if this is the right direction.
DOC-1921, spec here: https://docs.google.com/document/d/1TK98LD04jOImemlrWVKkL6AMufWzPDnlGVlWA7Rob6Y/edit#heading=h.98h83c2jnbei

### DIFF
--- a/en_us/install_operations/source/mobile.rst
+++ b/en_us/install_operations/source/mobile.rst
@@ -6,54 +6,213 @@
 Setting up the edX Mobile Applications
 ######################################
 
-This chapter is intended for those are who are building the edX mobile
-applications and customizing their Open edX installation to support their use.
+This section is intended for those are who are building edX mobile
+applications and customizing an Open edX installation to support their use.
 
-.. contents:: Chapter Contents:
+.. contents:: Section Contents 
+   :local:
+   :depth: 1
 
-****
-Code
-****
+******************************
+Accessing the Source Code
+******************************
 
-There are currently two edX mobile applications, one for iOS and one for Android. You can find the source code and additional documentation for each:
+There are currently two edX mobile applications, one for iOS and one for
+Android. You can find the source code and additional documentation for each
+application here.
 
-- iOS: http://github.com/edx/edx-app-ios
+* iOS: http://github.com/edx/edx-app-ios
 
-- Android http://github.com/edx/edx-app-android
+* Android: http://github.com/edx/edx-app-android
 
-**************
-Authentication
-**************
+*****************************************
+Configuring Mobile Application Features
+*****************************************
 
-The edX mobile apps require you to adjust your ``edx-platform`` configuration settings in ``lms.env.json`` to enable mobile API support.
+For the mobile API and authentication to work correctly with the edX mobile
+applications, you enable them in the ``lms.env.json`` file. You then complete
+the setup for authentication by creating OAuth clients and adding their IDs to
+the custom configuration file of each mobile application.
 
-In particular, you need to add the following under the features section:
+====================================
+Enable Mobile Application Features
+====================================
 
-.. code-block:: json
+.. note:: Configuration settings added to the ``lms.env.json`` file are reset 
+ to their default values when you use Ansible to update edx-platform. 
 
-    "FEATURES" : {
-        ...
-        "ENABLE_MOBILE_REST_API": true,
-        "ENABLE_OAUTH2_PROVIDER": true,
-        ...
+To enable the mobile application features, follow these steps. 
+
+#. In the ``edx/app/edxapp/lms.env.json`` file, add the following lines to the
+   features section.
+
+   .. code-block:: json
+
+       "FEATURES" : {
+           ...
+           "ENABLE_MOBILE_REST_API": true,
+           "ENABLE_OAUTH2_PROVIDER": true,
+           "ENABLE_COMBINED_LOGIN_REGISTRATION": true
+       }
+
+#. Save the ``edx/app/edxapp/lms.env.json`` file.
+
+#. Restart the server.
+
+.. future: add server- and client-side configuration for Google and Facebook sign in, incl. add ``"ENABLE_THIRD_PARTY_AUTH": true`` to the ``"FEATURES"`` list
+
+.. _Create the OAuth Clients:
+
+=======================================
+Create the OAuth Clients
+=======================================
+
+You create an OAuth client ID and secret that are specific to your
+installation for use by the mobile applications. The procedure that follows
+assumes that you create a different client for each of the edX mobile
+applications.
+
+Before you can create OAuth clients, a superuser must exist on the server. For
+more information, see `edX Managing the Full Stack`_.
+
+To create your OAuth clients, follow these steps.
+
+#. Log in to the Django administration console for your base URL. For example,
+   ``http://{your_URL}/admin``.
+
+#. In the **Oauth2** section, select **Clients**. 
+
+#. Select **Add client**. A dialog box opens with the **Client id** and
+   **Client secret** populated for the new client.
+
+#. Enter a **Url** and **Redirect Url** for the first application. For
+   example, ``https://{your_URL}/api/mobile/{version}/?app=ios``. While the
+   console requires values for these fields, they are not used by the edX
+   mobile applications. You can enter the same value in both fields.
+
+#. For the **Client type**, select **Public**.
+
+#. Optionally, enter a **User** and an identifying **Name**.
+
+#. Select **Save and add another**.
+
+#. Repeat steps 4-6 for the second application, and then select **Save**. 
+
+================================================
+Configure the Applications with OAuth Client IDs
+================================================
+
+The procedure that follows assumes that you have a different OAuth client ID
+for each edX mobile application.
+
+To configure each edX mobile application with its OAuth client ID, you add a setting
+to its corresponding custom configuration .yaml file. For information about
+how to set up custom configuration directories and files, see the GitHub
+repositories for `iOS`_ and `Android`_.
+
+#. Obtain the OAuth client id for the first application from your Django
+   administration console. For more information, see :ref:`Create the OAuth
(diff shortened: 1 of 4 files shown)


lamagnifica:
@aleffert, please look over the info I added for the app-side configuration per your docs.

aleffert:
@lamagnifica looks good :+1:

en_us/install_operations/source/mobile.rst
```diff
@@ -6,54 +6,201 @@
...
 applications and customizing their Open edX installation to support their use.
 
-.. contents:: Chapter Contents:
+.. contents:: Section Contents 
+   :local:
 
-****
-Code
-****
+.. Mark, this option ^ prevents the section name from appearing as a link. What do you think about standardizing on this format at the beginning of each rst?
```
mhoeber, line 15: not sure I understand, show me tomorrow

en_us/install_operations/source/mobile.rst
```diff
@@ -6,54 +6,201 @@
...
+=======================================
+Enable Features in ``server-vars.yml``
+=======================================
+
+You can enable many of the mobile application features in the 
+``server-vars.yml`` file. Configuration settings added to this file 
+are reapplied automatically when you update ``edx-platform``.
+   
+.. note:: Currently, you must manually add 
+          ``"OAUTH_ENFORCE_SECURE": " ",`` to the
```
mhoeber: ? should this value be blank?

mhoeber:
couple of questions, looks good. :+1:  

can you add change log entry too

en_us/install_operations/source/mobile.rst
lamagnifica: @nasthagiri or @jibsheet or @aleffert is this setting accurate?

jibsheet: This is likely a workaround for old ansible bugs.
I'm in the middle of turning it to false for all sandboxes, after which I think this line becomes irrelevant.
I'm running a quick test of setting it to false (Rather than "") on a newer sandbox right now.

jibsheet: Just tested, adding EDXAPP_OAUTH_ENFORCE_SECURE: False to server vars and it percolates down to lms.env.json properly.
This is probably still necessary on mobile3 because it has an outdated configuration, but will not be needed on newer mobile setups. @lamagnifica

lamagnifica:
@lou-wang please give your thoughts.

en_us/install_operations/source/mobile.rst
```diff
@@ -6,54 +6,201 @@
...
+following keys.
+
+::
+
+    PARSE:
+      NOTIFICATIONS_ENABLED: true
+      APPLICATION_ID: {your application id}
+      CLIENT_KEY: {your client key}
+
+
```
lamagnifica, line 185: per @lou-wang , no need to include the info in the remainder of this file

lamagnifica:
@nasthagiri completely rewritten after our meeting this afternoon. Please let me know if this is accurate, complete, and using terminology/phrasing that open edX developers are likely to understand. Thanks!

lamagnifica:
@jibsheet

en_us/install_operations/source/mobile.rst
```diff
@@ -6,54 +6,212 @@
...
+the custom configuration file of each mobile application.
 
-In particular, you need to add the following under the features section:
+====================================
+Enable Mobile Application Features
+====================================
 
-.. code-block:: json
+.. note:: Configuration settings added to the ``lms.env.json`` file are reset 
+ to their default values each time you update edx-platform.
```
nasthagiri: In case this needs to be more clear:
They are reset only each time you update a sandbox or a devstack vagrant image.
The `lms.env.json` file is not in the source code, so it doesn't get updated when the edx-platform source is updated.  For example, after a `git pull`.

en_us/install_operations/source/mobile.rst
```diff
@@ -6,54 +6,212 @@
...
 
-To configure a video module with a low resolution encoding for mobile, enter the URL to the mobile-targeted video as the first URL in the "Video File URLs" list in the video module's Advanced Editor in edX Studio.
+You create an OAuth client ID and secret that are specific to your
+installation for use by the mobile applications. The procedure that follows
+assumes that you create a different client for each of the edX mobile
+applications.
+
+To create your OAuth clients, follow these steps.
+
+#. Log in to the Django administration console for your base URL. For example,
```
nasthagiri, line 80: If they haven't already, they should create a super admin user on their server instance.
Readers can see `https://github.com/edx/configuration/wiki/edX-Managing-the-Full-Stack` for more information on managing an edX instance.

en_us/install_operations/source/mobile.rst
```diff
@@ -6,54 +6,212 @@
...
+   ``http://{your_URL}/admin``.
+
+#. In the **Oauth2** section, select **Clients**. 
+
+#. Select **Add client**. A dialog box opens with the **Client id** and
+   **Client secret** populated for the new client.
+
+#. Enter the **Url** and the **Redirect Url** for the first application. For example,
```
nasthagiri: You may want to make it clear that the `Url` and `Redirect Url` settings are not used by our code.  They can enter whatever value they want just for their own accounting purposes.

en_us/install_operations/source/mobile.rst
```diff
@@ -6,54 +6,212 @@
...
+
+   PARSE_KEYS = {

+     "APPLICATION_ID": "{app_id}",

+     "REST_API_KEY": "{API_key}"

+   }
+
+#. Save the ``edx/app/edxapp/cms.auth.json`` file.
```
nasthagiri: And restart the server after saving the updated file.

en_us/install_operations/source/mobile.rst
```diff
@@ -6,54 +6,212 @@
...
+
+#. In the **Contentstore** section, select **Push notification configs**. 
+
+#. Select **Add push notification config**. 
+
+#. Verify that **Enabled** is selected, and then select **Save**.
+
+#. Restart the server.
+
+.. is a restart needed?^^
```
nasthagiri: Restart is needed after saving the json file, but not needed after updating the Django admin settings.  This is why devOps prefers moving most of our configuration settings to Django admin so they don't have to restart/redeploy production servers.

nasthagiri:
:+1: lgtm overall, pending suggestions/comments above.

en_us/install_operations/source/mobile.rst
jibsheet: Possibly something like "are reset to their default values if you use ansible to update edx-platform" - this covers sandboxes, devstack and folks who run our plays (vs using git pull)

lamagnifica:
@jibsheet , @lou-wang

en_us/install_operations/source/mobile.rst
```diff
@@ -6,54 +6,213 @@
...
+           ...
+           "ENABLE_MOBILE_REST_API": true,
+           "ENABLE_OAUTH2_PROVIDER": true,
+           "ENABLE_COMBINED_LOGIN_REGISTRATION": true
+       }
 
+#. Save the ``edx/app/edxapp/lms.env.json`` file.
 
-Additionally, you need to create an OAuth key and secret specific to your installation for use by the apps. You can do this by logging into the Django administration console at ``<YOUR_EDX_INSTALLATION>/admin``. From there, choose **Clients** under the **Oauth2** section. If you don't already have any clients set up, you will need to add one. Choose **Add client** and create a client. The client id and secret that you see will need to be added to your mobile app configuration. For more information about how to do that, see the documentation for each app linked above.
+#. Restart the server.
```
jibsheet, line 60: If we always say "restart the server" and folks know we mean to restart a service, I'm fine with this, but I realize some people may reboot their devstack or instance which isn't needed.  They do need to restart the service (sorry I missed this on previous edits).  On a sandbox this would be something like

/edx/bin/supervisorctl restart edxapp:\* edxapp_worker:*

If we say "restart the server" elsewhere in the docs, then I'm fine with it.

en_us/install_operations/source/mobile.rst
```diff
@@ -6,54 +6,213 @@
...

+     "APPLICATION_ID": "{app_id}",

+     "REST_API_KEY": "{API_key}"

+   }
+
+2. Save the ``edx/app/edxapp/cms.auth.json`` file.
+
+#. Restart the server.
```
jibsheet, line 181: Same "restart the server vs service note" from above

jibsheet:
Overall looks good, but see my note about potentially confusing restart note.

en_us/install_operations/source/mobile.rst
lamagnifica, line 60: @jibsheet we say "restart the server" throughout the Installing & Configuring guide. See http://edx.readthedocs.org/projects/edx-installing-configuring-and-running/en/latest/.
That said, if they need to restart a service here instead of the server, I can easily rephrase this as "Restart the {name of service} service."

jibsheet, line 60: If "restart the server" is our convention, then ship it.  As a sysadmin I'm always sensitive to what requires an instance restart vs a service restart, but this isn't the place to go round and round about that.

jibsheet:
:+1:

PR #334, opened 2015-05-11
Add info about updating on-demand certificates

srpearce:


srpearce:
Merge when 5/12 release is live

PR #339, opened 2015-05-13
adds a reference to the website accessibility policy

cptvitamin:
@catong can you take a look at this minor change?  Now that we have an official policy, I wanted to reference it from this document, which many other documents are pointing to.


catong:
Tested the build locally and the link works fine. LGTM! :+1:

PR #343, opened 2015-05-15
DOC: Restructure Release Notes

catong:
This PR restructures release note files to allow indexing both by product and by date.
@mhoeber This is building without error for me locally. Can you please check before I merge? Thanks.


mhoeber:
@catong  this is great, thanks.  I made a few changes:
- in the product folders, the index files weren't used or needed; I removed those.
- you got build warnings for all the reuasable product/date files because they weren't included in any index. I added info at the end of the conf.py file to ignore those warnings, so we can better focus on real warnings.  Look at exclude patterns.  These use wildcards, so you don't have to keep adding to it if we keep the same naming standards.

:+1:  when build test finishes.  Thanks!

catong:
Squashed commits. Merging https://github.com/edx/edx-documentation/pull/343 when build is green.

catong:
Merging https://github.com/edx/edx-documentation/pull/343

PR #358, opened 2015-05-29
Ahodges/sd sfd assessments

lamagnifica:
@lou-wang @aleffert @nasthagiri this is the initial draft for the "Student Guide" mobile FAQ section. Se especially the Section_assessments.rst file.
@srpearce see include and only tags.

### DIFF
--- a/en_us/shared/students/SFD_mobile.rst
+++ b/en_us/shared/students/SFD_mobile.rst
@@ -4,117 +4,48 @@
 The edX Mobile App
 ########################
 
-The edX mobile app is a companion to the `edx.org`_ website. You can use the
-app to download course videos so that you can watch them whenever you want to,
-even without an Internet connection. To access the rest of the course,
-including course discussions, homework, and quizzes, you use a web browser.
-If you have access to a computer, you might find it easier to use it for some
-of the coursework.
+.. The following paragraph describes the features of the edX mobile app for partners/edx.org (video only)
+.. Alison, DOC-1840, June 2015
+
+.. only:: Partners
+
+  The edX mobile app is a companion to the `edx.org`_ website. You can use the
+  app to download course videos so that you can watch them whenever you want
+  to, even without an Internet connection. To access the rest of the course,
+  including course discussions, homework, and quizzes, you use a web browser on
+  a computer.
+
+.. The following paragraph describes the features of the edX mobile app for Open edX (adds notifications, assessments, discussions)
+.. Alison, DOC-1840, June 2015
+
+.. only:: Open_edX
+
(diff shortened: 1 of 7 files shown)


en_us/shared/students/Section_assessments.rst
```diff
@@ -0,0 +1,58 @@
...
+* Text input
+* Numerical input
+
+EdX is working to add more problem types to this list. 
+
+==================================================================
+Can I read all of the questions for my assignments in the app?
+==================================================================
+
+No. When you view the course in **full mode**, the courseware list identifies
```
aleffert: I don't think we use the "full mode" name within the app. The UI says something about "Full Course"

aleffert:
:+1:

en_us/shared/students/Section_assessments.rst
lamagnifica: got it. However, @aleffert , in the current Android build I am not seeing the expected indications for content that is not available on mobile. You navigate through multiple intervening pages before you get the message that it isn't available. Is that how we will deliver?

aleffert: Could you elaborate on the expected indications? I believe there are two things:
1. It gets a laptop icon in the course outline.
2. It shows that message when you actually try to go to the problem

It is possible that implementation isn't finished yet, though I think it is supposed to be.

cc @hanningni

lamagnifica:
@mhoeber this completes the partner-open split.
I built each version of the guide, and other than the section contents, the includes are working correctly in each one.

mhoeber:
@lamagnifica  not sure what you mean about section contents.  not working?

lamagnifica:
When a section contains one or more "include" statements, @mhoeber , the ".. contents::" command does not create links to their headings. The contents only has links to sections that actually appear in the same file. We saw this last week, so it's not unexpected.

mhoeber:
right ok

en_us/shared/students/SFD_mobile.rst
```diff
@@ -126,16 +57,35 @@ Courseware Questions
...
+
+.. The following paragraph describes the features of the edX mobile app for Open edX (adds notifications, assessments, discussions)
+.. Alison, DOC-1840, June 2015
+
+.. only:: Open_edX
+
+  Not entirely. With the edX mobile app, you can download course videos to
+  watch when you do not have an Internet connection. When you have an Internet
+  connection, you can also read course announcements and content, participate
+  in discussions, and do some, but not all, of the problems in your
```
BenjiLee: This is different from the information below about discussion forums.

BenjiLee:
:+1:

PR #365, opened 2015-06-02
June 2, 2015 Release Notes

mhoeber:
@srpearce @explorerleslie @jibsheet 

not a lot here.


en_us/release_notes/source/2015/discussions/disc_0602_2015.rst
```diff
@@ -0,0 +1,3 @@
+
+* When learners expanded an inline discussion, the bottom of the discussion
+  section was loaded. This problem is fixed. (TNL-2223)
```
explorerleslie: This is a little confusing the way that it is worded. What would happen is that if you expanded the inline discussion, the discussion would jump to some part of the discussion that was not the top. It's now fixed where the discussion stays at the top when expanded.

en_us/release_notes/source/2015/lms/lms_0602_2015.rst
```diff
@@ -0,0 +1,6 @@
+* A problem that sometimes caused the grade report to not complete, with no
```
explorerleslie: I wouldn't mention this in release notes. We fixed one of the issues that might cause this, but we're still seeing failing reports so there is still more work to do.

en_us/release_notes/source/2015/lms/lms_0602_2015.rst
```diff
@@ -0,0 +1,6 @@
...
+  error message, is fixed. (TNL-2251)
+
+* The calculator help window is updated for accessibility standards and now has
+  the ``aria-controls``, ``aria-haspopup``, and ``aria-describedby``
+  attributes. (AC-107)
```
srpearce, line 3: Probably not something for this week, but I wonder if eventually we should have a separate "Accessibility" section - we seem to be making accessibility updates every week, and this might help illustrate edX's commitment to accessibility.

srpearce:
:+1:

en_us/release_notes/source/2015/lms/lms_0602_2015.rst
mhoeber, line 3: @srpearce agree. next week?

srpearce, line 3: Sure!

PR #381, opened 2015-06-09
Adds section heads to the include directive, rather than the included file

lamagnifica:
@mhoeber. @srpearce , @catong also corrects some change log entries, some includes, and a renamed file. Please sync to this branch, build and make sure your recent changes are as you expect!


catong:
These changes look fine to me, though I don't understand why the headings shouldn't stay with the shared (included) topic but live in the "empty except for include" file instead. Does not having a top level heading cause an error in the shared file?

lamagnifica:
@catong, moving the 2nd level headings out of the included rst files and below the .. include:: directive instead means that they get picked up in the ToC that is automatically built in the left navigation frame and on the various index.html pages. 
- this is just a convenience
- this does not fix the problem with the contents:: directive (still doesn't include links to included files)
- no errors were caused by this, it's just a nice to have.

On the other hand, when I built the fours affected guides, make html threw a lot of errors. I fixed then as part of this same PR.

en_us/shared/students/SFD_wiki.rst
```diff
@@ -4,11 +4,10 @@
...
 
-Some courses include a wiki, which provides a public forum for both students and
-course staff to access, share, and collaboratively edit information about the
-course. If your course includes a wiki, you access it by selecting the **Wiki**
-tab at the top of any page. (If your course does not include a wiki, you don't
-see this tab.)
+Some courses include a wiki, which provides a public forum for both students
+and course staff to access, share, and collaboratively edit information about
+the course. If your course includes a wiki, you access it by selecting the
+**Wiki** tab at the top of any page. (If your course does not include a wiki, this tab is not present.)
```
srpearce, line 10: Nit: Line wrap

lamagnifica, line 10: not addressing the many line wrap errors in this wiki file at this time

srpearce:
One tiny, insignificant nit. Everything else checks out.

srpearce:
If you have a minute (and it's OK if you don't because these are my mistakes; I just thought this would be faster than creating another branch and PR) - I found a couple of broken links that need to be replaced.

In open_edx_course_authors/source/exercises_tools/notes.html
replace http://edx-guide-for-students.readthedocs.org/en/latest/SFD_notes.html
with http://edx.readthedocs.org/projects/open-edx-learner-guide/en/latest/SFD_notes.html

In open_edx_course_authors/source/building_course/search_course.html
replace http://draft-sfd-search.readthedocs.org/en/latest/SFD_course_search.html
with http://edx.readthedocs.org/projects/open-edx-learner-guide/en/latest/SFD_course_search.html

PR #383, opened 2015-06-09
Per request by Clayton, DOC-2052

lamagnifica:
@srpearce @catong @mhoeber might one of you have time to review?

### DIFF
--- a/en_us/video/source/front_matter/change_log.rst
+++ b/en_us/video/source/front_matter/change_log.rst
@@ -8,6 +8,9 @@ Change Log
 
    * - Date
      - Change
+   * - 11 Jun 2015
+     - Added the maximum video file length of 5GB to the :ref:`Specifications
+       for Successful Video Files` section.
    * - 22 Apr 2015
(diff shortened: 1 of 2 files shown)


catong:
@lamagnifica I can review.

catong:
LGTM :+1:  though I am curious as to whether the maximum is managed -- do users get an error if they try to upload a file > 5GB in size?

lamagnifica:
@catong this is the extent of the information that I have about this, also in the story:

Clayton Hainsworth
Jun 3 (6 days ago)

to docs, media, Lou 
Hi Docs,

Can I request that we add that the video upload tool has a 5 gig limit.

Thanks,
Clayton

I have forwarded your question on to Clayton. (I think it is highly unlikely that the message or status change, if there is one, would be returned in a useful time frame. It takes 24 hours for YouTube to turn around a file that is a couple of minutes long, and they warn on their own site that it is "possible" to upload videos larger than 2GB but that it will take a long time.)

PR #384, opened 2015-06-09
Remove bookmarks from Open edX student documentation

srpearce:
@catong, since you're here, can you give me some thumbs?


catong:
:+1:

PR #402, opened 2015-06-19
Fix build warnings

mhoeber:
@catong   @srpearce @lamagnifica 

Don't need all to review, but so everyone's aware.  made a few fixes for build errors.  Minor stuff. But we should be ensuring we don't introduce new warnings.  Guides build cleanly except for two warnings in the partners course staff project, warnings for links that are in conditions for open edx and not written to html file anyhow.  I'd like to figure out how to get rid of those warnings, for now we need to expect them.

### DIFF
--- a/en_us/shared/building_and_running_chapters/building_course/setting_up_student_view.rst
+++ b/en_us/shared/building_and_running_chapters/building_course/setting_up_student_view.rst
@@ -35,9 +35,9 @@ based on the content of the page.
 You configure the contents of this page in Studio, as described in this topic.
 
 .. note:: Given the diversity of MOOC learners, be careful to clearly 
-   communicate the  target audience, level, and prerequisites for your course. 
+   communicate the target audience, level, and prerequisites for your course. 
    Aim for concrete, unambiguous words (for example, ``understand eigenvalue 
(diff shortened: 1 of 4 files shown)


mhoeber:
Also, the moving of headings to include files doesn't make a difference in inline TOCs (need to fix too) but makes the file org cleaner

en_us/shared/building_and_running_chapters/building_course/setting_up_student_view.rst
```diff
@@ -35,9 +35,9 @@ based on the content of the page.
 You configure the contents of this page in Studio, as described in this topic.
 
 .. note:: Given the diversity of MOOC learners, be careful to clearly 
-   communicate the  target audience, level, and prerequisites for your course. 
+   communicate the target audience, level, and prerequisites for your course. 
    Aim for concrete, unambiguous words (for example, ``understand eigenvalue 
-   decomposition'' rather than ``intermediate linear algebra'').
+   decomposition`` rather than ``intermediate linear algebra``).
```
lamagnifica, line 41: good catch

lamagnifica:
agreed about the headings in section files
:+1:

PR #406, opened 2015-06-24
Fixes DOC-1983, Move video processing guide into the CA

lamagnifica:
@mhoeber please sanity check, you might want to make html for both open_edx_course_authors and course_authors... The "processing video files" section should now appear in the Partner B&R, but not in the Open B&R. Links have been adjusted to reflect intra-project locations.
The second commit includes a fix to an error on my part when pre-roll videos went out. (The include for the pre-roll video file should indicate that it is for partners only, and not for both audiences/guides.)

@lou-wang @catong @remyedX FYI, this will move the "processing video files" information out of a separate, private guide and into the partners' version of the Building & Running an edX Course guide.


mhoeber:
@lamagnifica   thanks.  Setup is fine.  If this is definitively just for partners, the files don't need to be in "shared", but it doesn't hurt anything that they are.  A few things:
- Change log
- Do we need to updated help config now to point to the new video index?
- Should we state more directly that this is not for Edge courses?
- Looks like there were warnings introduced by the accessibility update (grrr).  I will fix and you should rebase this and ensure there are no warnings.

lamagnifica:
@mhoeber, comments addressed and branch rebased. no unexpected errors during make html

PR #419, opened 2015-07-01
Fixes DOC-736, add a workaround for transcript with special characters

lamagnifica:
@JAAkana , @srpearce @mhoeber @catong , this is for https://openedx.atlassian.net/browse/DOC-736. I could not myself get the Advanced page workaround to work, so I only provided only Jennifer's original workaround instead of both.


mhoeber:
:+1:

PR #423, opened 2015-07-01
Archive older change log without links

mhoeber:
@lamagnifica  @catong  @srpearce 

Moved older change log to separate file and removed links so we don't need to always maintain them.


en_us/course_authors/source/front_matter/change_log_2014.rst
```diff
@@ -0,0 +1,596 @@
+.. _Change Log 2014:
+
+################################
+Change Log for 2013 through 2014
+################################
+
+This section lists changes in this document from 2013 and 2014. 
+
+For more recent changes, see the :ref:`Change Log for 2015`.
```
srpearce: Strike "the" or make "change log" lowercase

srpearce:
One nit - otherwise :+1:

PR #429, opened 2015-07-06
Document ability to preview LMS as a particular student, reset student attempts for problems from RCBs

catong:
Document course team member ability to preview the LMS as a particular student and reset student attempts for specific problems that a student was assigned from an RCB.
(Combines SOL-858, SOL-816 -> DOC-2120)  https://openedx.atlassian.net/browse/SOL-1027, https://openedx.atlassian.net/browse/SOL-1028)
@antoviaque @smarnach  @pbaruah - please review for factual accuracy vs implementation. Thanks!
@griffresch Just FYI


en_us/shared/building_and_running_chapters/developing_course/testing_courseware.rst
```diff
@@ -32,16 +33,24 @@ You can preview course content before a course is live or before you publish
...
+   information about previewing course content if you use content groups to
+   designate content as visible only to particular groups, see :ref:`Preview
+   Cohort Specific Courseware`.
+
+.. note:: If you use randomized content blocks in your courseware, you cannot
+   preview unpublished units that contain content from randomized content
+   blocks, because the randomized content is not assigned until after the unit
+   is published. For information about viewing the actual content that has
+   been assigned to a particular student from a randomized content block in a
+   live course, see :ref:`Specific Student View`.
```
catong, line 54: @smarnach @antoviaque Before a course is live, I'm not clear what is seen in the LMS for randomized content block content when a course team member selects "View as **Student**" or "View as "Specific Student".
Do they see the ALL possible matching content in the RCB, or matching RCB content limited by the number that is specified to be assigned to one student? Is there any difference in that situation between viewing as "all students" or "one particular student"? My understanding is that the content is not assigned at least until the unit is published (or until it is published and a user accesses that unit?) so are we showing what "all students" might see from the RCB?

en_us/shared/building_and_running_chapters/developing_course/testing_courseware.rst
```diff
@@ -174,6 +214,47 @@ In **Student View**:
...
+  :ref:`Published and Live`. For units that are
+  :ref:`Draft (Unpublished Changes)`, you see the last published version of the
+  unit. 
+
+* You do not see units that are :ref:`Draft (Never Published)` or
+  :ref:`Visible to Staff Only`. To see these units, you must switch back to Staff view or use Preview mode as described in :ref:`Preview Course Content`.
+
+* You can access the Instructor Dashboard, which has features and reports that
+  help you run your course.
+
```
catong, line 264: When a course team member masquerades as a student who is a member of a cohort, the preview/live view does take into account cohort-specific courseware, is this correct?
@antoviaque @smarnach

en_us/shared/building_and_running_chapters/developing_course/testing_courseware.rst
antoviaque, line 54: @catong I believe in this case it would behaves the same way as for blocks that students haven't visited yet - ie it will initialize the ContentBlock and make a random selection from the library, but since the changes aren't recorded, this selection will be temporary. When the user does end up visiting the LibraryContent block (or if the instructor reloads the "view as specific student" view), it will still be uninitialised, and another set of blocks from the library will be selected.

See http://sandbox4.opencraft.com/courses/course-v1:OpenCraftX+1+2015/courseware/b5e4cfc740444c2f8a0121412ed137ff/0d3fe13346f146d294c169fd90296cdb/ for a demonstration of this (login as staff and masquerade as "honor", which currently hasn't visited this new section)

en_us/shared/building_and_running_chapters/developing_course/testing_courseware.rst
antoviaque, line 264: @catong From my tests, it doesn't at the moment -- it would make sense that it does behave this way, but it wasn't within the scope of the current task, so we haven't added explicit support for this. It would however be a good follow-up task, as I can see cases where this would be useful.

Example: http://sandbox4.opencraft.com/courses/course-v1:OpenCraftX+1+2015/courseware/b5e4cfc740444c2f8a0121412ed137ff/dbdcc9b9f6d24a518d92e4efbabd986e/ (login as staff, and view as "honor" - honor is in the group/cohort 2)

antoviaque:
Otherwise looks good!

One thing I noted is that we don't seem to warn staff about the fact that changes that they make in that mode won't affect the students - unless I missed a piece neither the UX or the documentation mention this. You don't think this will be confusing for users, that they would expect the changes they make to a given student exercises' to be saved?

en_us/shared/building_and_running_chapters/developing_course/testing_courseware.rst
catong, line 54: @antoviaque thanks for the link. If I understand correctly what I'm seeing from your test course, all of the "View as..." options see a filtered view of the RCB content -- that is, if the RCB is supposed to return 2 problems, a user in any of the "View As" roles, (including "Staff"), sees 2 problems in the LMS, even if these are not the final assigned problems. 

In my experience with your test course, the set of returned problems shown to each user in the preview conditions remained the same after refreshing or switching users. Is this because as soon as there is a "view/preview" by a user it is considered initialized? 

Also, and only for my education, what is the reason why Preview does not work in the sandbox? That particular area of code not included, or some other reason?

en_us/shared/building_and_running_chapters/developing_course/testing_courseware.rst
catong, line 264: Thanks, @antoviaque I did some testing using my own course with cohorts. What I'm observing is that when masquerading as a specific student, I cannot see the content from an RCB that I DO receive as a member of a content group. Is that also your understanding of the behavior when you say "no explicit support" for cohorts? 
I would expect that even if masquerading as a specific student does not completely duplicate the course experience in that some content is NOT hidden from me when it should be, I should at least expect that I should see content that IS supposed to be assigned to me. 
Can you please confirm? 
http://sandbox4.opencraft.com:18010/course/course-v1:CTX+CT200+2015_T2?show=block-v1%3ACTX%2BCT200%2B2015_T2%2Btype%40sequential%2Bblock%4056f60470149e4d4c8b68d28fc03b0aa4

I _think_ I am seeing at least a bug in that no content can be viewed live or previewed for a specific student as soon as they are in a cohort, but I'm also a bit concerned about how to present the limitations of this feature to course teams. @pbaruah and @explorerleslie can you please advise?

catong:
@pbaruah @explorerleslie @antoviaque 
Upon further testing, I still observe the behavior that I can't see any course content when masquerading as a student in a cohort, even if the unit does not contain any content from an RCB.
Please see the 2nd subsection in my test course, which contains all non-RCB content. Two units, one visible to Content Group A (audit@example.com) one visible to Content Group H (honor@example.com).
http://sandbox4.opencraft.com/courses/course-v1:CTX+CT200+2015_T2/courseware/c5d428de064f46bfa657ed59e6e7fcc5/17be58e076a5482d98ab777816ad7388/

explorerleslie:
I discussed this with @pbaruah, but documenting here as well. I'm not comfortable releasing this feature if it's not truly "preview as a specific student", including previewing content from content groups and experiment groups.

en_us/shared/building_and_running_chapters/developing_course/testing_courseware.rst
bradenmacdonald, line 54: @catong When you switch users it should sometimes show different content - the "staff", "student", and "specific student" views will all show a filtered view of the RCB content, but each one should be a different (random) filtered view. In this case, "honor" and "staff" do show different content. When staff views the course as honor, they will always see exactly the content that honor will see - the only exception would be if honor has not yet visited that content, in which case the random filtering has not yet been saved for the honor user.

The Preview link in Studio _does_ seem to be working on the sandbox when I tried it just now so I'm not sure about your last question. I'll let Xavier chime in if there's something else going on.

bradenmacdonald:
@catong I just checked into that issue. You are right, but there is a slight clarification needed: When masquerading as a user, cohorted content is not visible, but non-cohorted content still appears fine. So it's not correct to say that one "can't see any course content when masquerading as a student in a cohort". Rather, one can't see cohorted content when masquerading.

en_us/shared/building_and_running_chapters/developing_course/testing_courseware.rst
antoviaque, line 54: @catong @bradenmacdonald Yup, the content can change randomly - there isn't a lot of blocks in that library, so sometimes the selection can remain the same, or very similar, even if the dices have been rolled again.

For the preview mode, it doesn't work there (you can click the preview button in Studio, but it just brings you to the "normal" LMS). I'm not sure why the standard fullstack deployment (or devstack) don't include the preview mode, but I'm guessing that this might be either because no one bothered about it yet, or maybe because of memory constraints (the preview mode is actually another instance of the LMS, with a slightly different configuration, if I remember correctly - so that increases the memory footprint). We haven't needed it much so far, so I didn't look into it to add it to our sandboxes, but if we start needing it more, I'll look into it.

en_us/shared/building_and_running_chapters/developing_course/testing_courseware.rst
antoviaque, line 264: @catong From my tests, you do see the content that is visible from all cohorts, but not the content that is restricted to a specific cohort - ie a way to describe this would be that content which had their visibility setting edited, to be only visible from a portion of the users, wouldn't be visible in the "view as specific user" mode. Does that match what you have seen?

If we do want to support this, we can definitely add a new story about this.

antoviaque:
Oh, just saw these additional comments after replying to the inline thread - +1 to what Braden mentioned. We can add explicit support for cohorted content, but this is likely to take time to implement, and we still need the current feature for the MIT course starting at the end of the month. If we delay the release of this feature for it, it's going to bring its release dangerously close to the start of the course. 

@explorerleslie If the precisions about what is currently excluded help, and you agree with us that the current feature does stand on its own, it would be better to be able keep the current scheduled release while we work on the additional scope - that would help ensuring the base feature is safe, and that we have time to get bug reports & address them before the course starts.

Edit: A precision - the feature is currently slated for the release cut at the end of the week, not the one currently being deployed.

en_us/shared/building_and_running_chapters/developing_course/testing_courseware.rst
catong, line 264: @antoviaque Yes, that matches what I see.

pbaruah:
@catong @antoviaque @explorerleslie @bradenmacdonald 

After going through verification on the sandbox, discussions with many of you, this is what I think we will be doing -

The new functionality that has been delivered by this PR covering SOL-816 are:
- Ability for instructor to preview courseware as a specific student
- Ability for instructor to reset specific RCB question (e.g. question # 2 out of a set of 10)
- Ability for instructor to reset the whole RCB (student gets a new set of 10 questions)
- Ability for instructor to reset specific non-RCB question (e.g. question # 3 in assignment 5)

During verification, it was noticed that "Preview as student" (or the new "Preview as a specific student") does not work for cohorted/experiments content whether it is RCB or regular content. There is an existing PR/bug -

https://github.com/edx/edx-platform/pull/4460
https://openedx.atlassian.net/browse/TNL-377

We will keep this separate from the current PR and Leslie and I can discuss a schedule to fix this.

explorerleslie:
@pbaruah the Preview As Student functionality is meant to work as a general preview from the student perspective and is explicitly meant to exclude cohorted and experiment content, because there is no filter to say what student you're looking at and whether they have any special content. 

However, I still feel that if you're messaging and selling this as Preview as Specific Student, the user's expectation is that you will see the unit _exactly_ as the student sees it, which includes cohorted and experiment content. Let's talk more if you disagree.

catong:
@antoviaque @pbaruah @explorerleslie 
I'd like to wrap up the documentation in a state that it is consistent with the code in master at this time, just in case that is released, and also so that the doc reviews for SOL-816 and SOL-858 can be considered completed.
My plan is to add a note qualifying this "preview as specific student" feature. This note will indicate that if the course contains any content that is visible only to a specific group (either content group or experiment group), then that content cannot be previewed using the "View as specific student" feature.
Then, as SOL-1064 is completed, I will update this documentation to reflect those functionality changes.
Can you please confirm that this is OK, or provide corrections/recommendations? Thanks.

antoviaque:
@catong Replied yesterday in the chat, but confirming here for reference - that seems like a good approach to me, yes. Thank you!

catong:
Squashing commits. Considering this PR "doc complete". Will either release as is with SOL-858 and SOL-816 code, or will be modified to remove the qualifying notes if SOL-1046 is finished in time for the release.

catong:
Merging https://github.com/edx/edx-documentation/pull/429

catong:
For the record, thumbs were given in comments on a commit: https://github.com/edx/edx-documentation/commit/12ae0ee9deaf6ba50215ebe66458bf78427d22c1#commitcomment-12237965

PR #467, opened 2015-08-06
Fixes numbering in new include section

lamagnifica:
@mhoeber @srpearce @catong if one of you would do the honors... I realized that the numbering for this include is incorrect not just locally with make html, but in the RTD built files as well (See http://edx-partner-course-staff.readthedocs.org/en/latest/getting_started/accounts.html#create-an-account). Fixes the numbering.


srpearce:
:+1:

PR #468, opened 2015-08-06
Insights release of total enrollment

lamagnifica:
@shnayder @dsjen release notes!
@srpearce @catong @mhoeber this feature should be releasing today...

### DIFF
--- a/en_us/release_notes/source/2015/insights/insights_0806_2015.rst
+++ b/en_us/release_notes/source/2015/insights/insights_0806_2015.rst
@@ -1,0 +1,10 @@
+
+A new metric that measures total interest in your course over time is now
+available in edX Insights. The **Enrollment Activity** page now includes a
+metric for **Total Enrollment** (the number of learners who have ever been
+enrolled in the course) as well as for **Current Enrollment** (the number of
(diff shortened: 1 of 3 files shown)


dsjen:
LGTM! :dancers:

en_us/release_notes/source/2015/insights/insights_0806_2015.rst
```diff
@@ -0,0 +1,10 @@
+
+A new metric that measures total interest in your course over time is now
+available in edX Insights. The Enrollment Activity page now includes a metric
```
srpearce: Bold "Enrollment Activity"

srpearce:
One nit - otherwise :+1:

mhoeber:
Seems potentially confusing that one could see change listed as 8/6 and 8/4.  otherwise :+1:

lamagnifica:
@mhoeber I agree, but hoped that the change from "as of" to "week of" would make that less confusing. How should we word these and/or date these going forward, since multiple releases per week (and not just to platform) will become more and more the norm?

mhoeber:
@lamagnifica   yes I see.  I think it's fine and a pattern we can keep

PR #470, opened 2015-08-06
Further reduces prominence of ORA 1 for events

lamagnifica:
@explorerleslie I've been meaning to do this for a while. Strips the ORA1 events out of an alphabetized list of reference links, and provides date deprecated as well as completely removed. DOC-430
@catong or @srpearce if you'd be so kind as to check the changes for compliance...


explorerleslie:
:+1:

PR #484, opened 2015-08-12
Adds third party authentication announcement

lamagnifica:
@catong please check!


en_us/release_notes/source/2015/website/website_0812_2015.rst
```diff
@@ -0,0 +1,19 @@
+
+==========================================
+Third Party Authentication with edX Edge
+==========================================
+
+To enhance sign in options and streamline access to courses running on the edX
```
catong, line 6: I actually preferred the "To provide additional..." version rather than enhanced... seems more precise? But up to you.
Also the "sign in" vs "sign on" issue. You have "sign in" here but "sign on" in the portal announcement.

catong:
Minor question/nits, otherwise :+1:

en_us/release_notes/source/2015/website/website_0812_2015.rst
lamagnifica, line 6: hmm, this is the one Beth approved, I tried to make it a little softer and squishier for the portal. 
I have no idea what to suggest for sign in vs. sign on, since "sso" is a thing. It looks like we've painted ourselves into a corner

PR #488, opened 2015-08-12
Updated to include separate discussion info for partners and open edx

lamagnifica:
Also updates text and an image for DOC-2062 (section download no longer
supported), and removes an out of date and redundant image

@lou-wang, @catong, @aleffert these are pretty lightweight updates to the Learner's guides. Delivery requested by 17 Aug.


aleffert:
Those arrow icons should be right aligned. That seems like a bug (and probably not an image we want in our docs)

aleffert:
Are we ready to replace that list image? It's not released for the edx.org version of the app.

aleffert:
Otherwise, :+1:

lamagnifica:
yeah, I assumed that the image had some problems, and tried to align the icons myself -- clearly not well enough.
you're also right that I didn't consider that there are different images needed (if we include them) for partners and for open edX. Let me pnder.
Thanks, @aleffert

lamagnifica:
@aleffert I addressed one issue by adding the original image back, and making the images audience-specific. For the other, since we might be delivering doc before sw, I don't think it's crucial to have a final image right now, so noted that we'll probably need a replacement.

aleffert:
To be clear, about the right alignment I meant something more like this 
mob_downloadicon_openedx

lamagnifica:
@aleffert yes, understood. Would rather not spend a while lot of time tweaking images by hand though.

en_us/shared/students/SFD_mobile.rst
```diff
@@ -90,10 +90,24 @@ Can I take a course entirely on my mobile device?
...
 
-Right now, you cannot use the app to participate in course discussions. To
-read or contribute to the discussions, use a web browser on a computer.
+.. The following paragraph describes the features of the edX mobile app for partners/edx.org
+.. Alison, DOC-1840, June 2015
+
+.. only:: Partners
 
-.. Reviewers, I will provide the Open_edX version of this paragraph ^^ in a different PR - Alison
+  Right now, you cannot use the edX mobile apps to participate in course
```
catong, line 99: "app" singular?

en_us/shared/students/SFD_mobile.rst
```diff
@@ -90,10 +90,24 @@ Can I take a course entirely on my mobile device?
...
+  discussions. To read or contribute to the discussions, use a web browser on
+  a computer.
+
+.. The following paragraph describes the features of the edX mobile app for Open edX (adds notifications, assessments, discussions)
+.. Alison, DOC-1840, June 2015
+
+.. only:: Open_edX
+
+  You can read and contribute to course discussions in the edX mobile apps
```
catong, line 108: "apps" -> "app"?

en_us/shared/students/SFD_mobile.rst
```diff
@@ -131,31 +162,18 @@ if you intend to watch or download a lot of videos. The edX mobile app has a
...
-
-========================================================
-What are the circles to the left of the video name?
-========================================================
-
-The circle indicates whether you have already played a video. A blue circle
-means that you have not started playing the video. A half-filled circle means
-that you played part of the video. A gray circle means that you played the
-entire video.
+The number to the right of a subsection name is the number of videos in that
```
catong: "is the number" -> "indicates the number of videos"

en_us/shared/students/SFD_mobile.rst
```diff
@@ -131,31 +162,18 @@ if you intend to watch or download a lot of videos. The edX mobile app has a
...
+subsection.
 
 ========================================
 How can I make the video full screen?
 ========================================
 
 Rotate your phone until the video changes to landscape, or select "full
```
catong: "landscape" -> "landscape mode"

en_us/shared/students/SFD_mobile.rst
```diff
@@ -169,8 +187,7 @@ You can change video speeds on the iPhone app. While a video plays, select
 **Settings** (the "gear" icon) at the lower right, and then select **Video
 Speed**.
 
-On the Android app, videos currently run only in their original speeds. EdX is
-working on the ability to change video speeds.
+On the Android app, videos currently run only in their original speeds.
```
catong: "run" is a bit odd. "play" instead? and "at" a speed, not "in" a speed.

"videos currently play only at their original speed."

en_us/shared/students/SFD_mobile.rst
```diff
@@ -210,6 +227,16 @@ How do I delete downloaded videos?
 
 #. Select **Delete**.
 
+========================================================
+What are the circles to the left of the video name?
```
catong: Is it one circle per video? In that case "What is the circle to the left of the video name?" If you want to keep things plural, "What are the circles to the right of video names?"

catong:
:+1:  with a couple of minor edits.

en_us/shared/students/SFD_mobile.rst
lamagnifica, line 99: Turns out that referring to our "app" in the singular is a bit of a misnomer. There are two apps, one that runs on Android smartphones and one that runs on iPhones. They are completely different, though the engineering and UX teams have worked really hard to make the user interfaces as similar as possible. It's a bit of a struggle to get every reference in the doc correct though.

en_us/shared/students/SFD_mobile.rst
lamagnifica: thank you for making suggestions on the original text -- I agree with all of them, though I have not been prioritizing rewriting existing content. I will try to make as many of them as possible.

PR #489, opened 2015-08-13
Adds what won't work in the app for open edX users only

lamagnifica:
@aleffert @lou-wang @awaldrop please review for accuracy & completeness.

### DIFF
--- a/en_us/open_edx_course_authors/source/front_matter/change_log.rst
+++ b/en_us/open_edx_course_authors/source/front_matter/change_log.rst
@@ -12,6 +12,10 @@ August 2015
 
    * - Date
      - Change
+   * - 17 August 2015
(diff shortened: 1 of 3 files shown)


aleffert:
:+1:

lamagnifica:
@lou-wang and @andywaldrop please review the commit above (https://github.com/edx/edx-documentation/commit/b9549e7f21e415c26a21a64ae3da26a41c02d7d0) if you have a chance to take a look. I'll be adding a commit that addresses style issues and brings these files into line with our style guide, but there's no need for you to wade through those changes.

lamagnifica:
@catong the latest commit adds changes for some of the style guide issues that I found. This needs to go on Monday, so if you prefer, you can just look at the feature-related changes in the previous commit as I've suggested to Lou and Andy.

PR #506, opened 2015-08-20
Removes the VitalSource tool from the shared directory

lamagnifica:
Completes DOC-928, which requested removal of the VitalSource tool. 
- Deletes vitalsource.rst file (now attached to DOC-2227 iF ever needed).
- The referenced image was removed months ago.
- Deletes the obsolete & unused index.rst file from the shared exercises_tools directory. No other directories under shared have index.rst files.

@mhoeber @catong @srpearce , burning it with fire.


mhoeber:
:+1:

PR #509, opened 2015-08-22
openassessment event titles

clankjak:
openassessment.create_submission
openassessment.save_submission
openassessment.student_training_assess_example
openassessment.upload_file

should be openassessmentblock

PR #516, opened 2015-08-26
Corrects the names of two ORA2 events

lamagnifica:
Based on the work of @clankjak in https://github.com/edx/edx-documentation/pull/509.
@stroilova if you would be so kind as to triple-check?
@catong or @srpearce when you have a moment for a quick sanity check.


stroilova:
:+1:  Thank you.

catong:
@lamagnifica Looks fine to me. :+1:

PR #517, opened 2015-08-26
DOC-1954 Glossary additions

catong:
https://openedx.atlassian.net/browse/DOC-1954
Add "whitelist" and "embargo" to the glossary as they are used in explaining what values might occur in what situation in the "Certificate Eligible" column of the grade report.
@explorerleslie Can you take a quick first look, please? Thanks!

### DIFF
--- a/en_us/shared/glossary/glossary.rst
+++ b/en_us/shared/glossary/glossary.rst
@@ -308,6 +308,17 @@ E
 
   For more information, see `What is Studio?`_.
 
+.. _embargo:
+
+**Embargo**
+
+  An embargo is an official ban on trade or commercial activity with a
+  particular country. For example, due to U.S. federal regulations, edX cannot
+  offer certain courses (for example, particular advanced STEM courses) on the
+  edx.org website to learners in embargoed countries. Learners cannot access
+  restricted courses from an embargoed country. In some cases, depending on the
+  terms of the embargo, learners cannot access any edX courses at all.
+
 
 .. _Exercises:
 
@@ -816,6 +827,21 @@ V
 W
 ****
 
+.. _Whitelist:
+
+**Whitelist**
+
+  In edX courses, a whitelist is a list of learners who are being provided with
+  a particular privilege. For example, whitelisted learners can be specified as
+  being eligible to receive a certificate in a course, regardless of whether
+  they would otherwise have qualified based on their grade.
+
+  In the grade report for a course, whitelisted learners have a value of "Yes"
+  in the **Certificate Eligible** column, regardless of the grades they
+  attained. For information about the grade report, see `Interpret the Grade
+  Report`_.
+
+
 .. _Wiki:
 
 **Wiki**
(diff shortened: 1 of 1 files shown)


en_us/shared/glossary/glossary.rst
```diff
@@ -297,6 +297,20 @@ E
...
 
+.. _embargo:
+
+**Embargo**
+
+  Embargoed countries are those that are subject to economic or trade
+  sanctions. Due to U.S. federal regulations, edX cannot offer certain courses
+  (such as particular advanced STEM courses) to learners in embargoed
+  countries. Learners are not able to access restricted courses from an
+  embargoed country.
```
explorerleslie: There's actually another case, where all courses are blocked in a particular country, see: http://blog.edx.org/education-right-not-service-new/

catong:
@lamagnifica @srpearce @pdesjardins - can one of you do a quick review? 
Any recommendations for additional reviewers, or go with Leslie's previous review + doc?

en_us/shared/glossary/glossary.rst
```diff
@@ -816,6 +826,20 @@ V
 W
 ****
 
+.. _Whitelist:
+
+**Whitelist**
+
```
lamagnifica, line 833: I'd be a little more general with this -- whitelist is a jargon term that can be used for many features. Perhaps characterize the certs info as an example after providing a more generic description? (I avoided using "blacklist" for the ORA file types feature in the B&R guide, but used it in the ICR guide...)

en_us/shared/glossary/glossary.rst
```diff
@@ -308,6 +308,16 @@ E
 
   For more information, see `What is Studio?`_.
 
+.. _embargo:
+
+**Embargo**
+
+  Embargoed countries are those that are subject to economic or trade
```
lamagnifica: perhaps start with what an embargo is, and then go on to embargoed countries? or, change the term being defined to "Embargoed Countries"?

en_us/shared/glossary/glossary.rst
```diff
@@ -308,6 +308,16 @@ E
...
+  sanctions. Due to U.S. federal regulations, edX cannot offer certain courses
```
lamagnifica: courses ^on the edx.org website

en_us/shared/glossary/glossary.rst
```diff
@@ -308,6 +308,16 @@ E
...
+  (such as particular advanced STEM courses) to learners in embargoed
+  countries. Learners cannot access restricted courses, or in some cases, any
```
lamagnifica: "any" here seems to contradict the "particular STEM" statement immediately above

lamagnifica:
prob more suggestions than needed, :+1:

catong:
@lamagnifica You already gave thumbs but might want to take a final look.

lamagnifica:
:+1: :+1:

PR #527, opened 2015-09-03
Correct description of how Edge and edx-lti.org interact

lamagnifica:
@jibsheet and @srpearce please take a quick look at this correction for me. 
The doc should not give the impression that edx-lti.org acts as the host site for any courses. Either Edge or an open source instance set up by the partner org will be the host.


srpearce:
:+1:

jibsheet:
:+1:

PR #539, opened 2015-09-08
Upgrade Travis to use new container infrastructure

sarina:
See http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
+sudo: false
 language: python
 branches:
     only:
         - master
 python: "2.7"
+cache: pip
 # This part isn't needed until we are running latexpdf builds.
 # before_install:
     # MacTex is used locally (MacOSX) by the docs team,
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,23 @@
 sudo: false
 language: python
-branches:
-    only:
-        - master
-python: "2.7"
-cache: pip
-# This part isn't needed until we are running latexpdf builds.
-# before_install:
-    # MacTex is used locally (MacOSX) by the docs team,
-    # but since we are using a linux machine for travis,
-    # we can use texlive-full instead.
-    # - "sudo apt-get update"
-    # - "sudo apt-get install texlive-full git python-dev -y"
+python:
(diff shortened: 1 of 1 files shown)


sarina:
@singingwolfboy or @jzoldak or @benpatterson maybe?

.travis.yml
```diff
@@ -1,8 +1,10 @@
+sudo: false
 language: python
 branches:
     only:
         - master
 python: "2.7"
+cache: pip
```
benpatterson: Unfo this won't work. See example [here](https://github.com/edx/ecommerce/blob/master/.travis.yml#L6).

And, a more concise example of a workaround is [here](https://github.com/semenko/https-everywhere/commit/1e17430a57a9bc8fe048757a4c03bf5dbe0479d1).

sarina: Ugh, really? we've got a ton of repos to fix, then :cry:

.travis.yml
```diff
@@ -1,8 +1,12 @@
+sudo: false
 language: python
 branches:
     only:
         - master
 python: "2.7"
+cache:
+    directories:
+        - $HOME/.cache/pip
```
sarina, line 7: @benpatterson this better?

jzoldak:
@clintonb is making a [cookiecutter django template repo](https://github.com/clintonb/cookiecutter-django-edx/pull/1/files#diff-2c540bfcb102e51fbe600b5b48f35f0e).
Should we standardize on an order that we put these lines in?

sarina:
I don't understand how the django template repo would relate to non Django based repos.

But yes, standardizing the order makes sense. However if it's going to be a lot of debate I'd like to just merge this and let that discussion happen elsewhere.

clintonb:
@sarina @jzoldak I have no preference on order for `.travis.yml`. It's a file I rarely open.

sarina:
I put the file in the same order as i18n-tools and other repos I use

jzoldak:
:+1:

PR #549, opened 2015-09-13
Fixes DOC-2017

lamagnifica:
Removed an outdated image, attempted to make the process clearer, and applied new style guidelines to the section on finding a location identifier. The interesting part starts at line 541.
Per @JAAkana 's request in [DOC-2017](https://openedx.atlassian.net/browse/DOC-2017)
This turns out to be similar, but not quite the same, as a procedure that course teams need to do to identify components for use in an LTI situation, written last week...
@mhoeber @catong @srpearce @sstack22


srpearce:
:+1:

sstack22:
:+1:

PR #559, opened 2015-09-18
Fix description of ECOM-2118 in RN

catong:
In RN for Sept 16, the description for the fix for ECOM-2118 should have described sorting by start date rather than due date.


catong:
Merging https://github.com/edx/edx-documentation/pull/559

PR #560, opened 2015-09-18
Use Intersphinx for XBlock Tutorial

mhoeber:
This is a PR to use Intersphinx for the first doc project.

Notes:
- There is a new shared config file, at the top level:  shared/conf.py.  The project conf.py file in Xblock tutorial imports most of its settings from this. We will update other projects to do this as well, as most settings are changed. That way we can make changes in one place.
- The master conf.py file defines the intersphinx projects -- see lines at bottom.  These define the tokens and their related projects.  Those tokens are then used in individual links.
- Links to other doc projects are changed to use the syntax :ref:`intersphinx-token-for-project:label`
  This way, we never use the URL to doc; if it changes the link will be ok (as long as we don't change the labels. If we are linking to a topic without a label, we need to add that label.
- Links to non-doc sites, for example Github, still need to be in a links.rst file.  I put these for XBlock project in a top level file, links/links.rst.  We should get all links in there and shared by all doc, so we don't have to duplicate and maintain.

@lamagnifica @catong @srpearce   process ^^^ and take a look at files.  We can go over more thoroughly in meeting next week.


lamagnifica:
My $.02 on your 3rd bullet, if I understand it correctly, is that it would be very desirable to reduce the number of different labels that we have for one guide: course_authors is the name of the containing directory, edx-Partner-Course-Staff is the name of the RTD project, Partners is the name of the only:: tag... I'd rather that we did not now add partnercoursestaff to indicate the same content in a different context. We could make a sweeping change and get them all in line, of course, or at least maybe reuse one of them instead of adding a new one.

mhoeber:
@lamagnifica   I agree . . . but for these labels for intersphinx -- annoyingly, no non alpha characters, no upper case.  only works with all lowercase letter strings.

PR #583, opened 2015-10-01
Make file tests all projects

mhoeber:
@lamagnifica @catong @srpearce 

This is an updated make file at top level of en_us.

This DELETES build folders and regenerates each build.

We've got to get to point where we run this at completion of any PR and get 0 warnings.  There are a handful of warnings in here now from intersphinx work. I will fix those and merge in another PR before merging this one.

All you need to do is go to en_us direction enter enter **make**

### DIFF
--- a/en_us/Makefile
+++ b/en_us/Makefile
@@ -1,5 +1,5 @@
 
-.PHONY: html
+.PHONY: all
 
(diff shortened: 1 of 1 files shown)


lamagnifica:
:+1:

PR #585, opened 2015-10-02
Fix all intersphinx issues

mhoeber:
@lamagnifica   this fixes all remaining intersphinx issues.

Run make for all, and there is one warning -- for doc link to research guide in release notes.  This PR will fix this, but research guide will need to be rebuilt after this PR is merged to get rid of warning.  You shouldn't get any other warnings.


lamagnifica:
As expected, there's a single error when building hte Release notes involving the research guide. :+1:

PR #586, opened 2015-10-05
Removed caution about updating on-demand certs

mhoeber:


mhoeber:
@griffresch or @mattdrayer   this is from bug Julia submitted

mhoeber:
@mattdrayer  I updated based on testing.  Support says they currently can't update scores for people.  I'm following up with Julia.

mhoeber:
@mattdrayer  updated, can I get a thumb

PR #595, opened 2015-10-07
Fix install guide links

mhoeber:
@lamagnifica


lamagnifica:
make runs without errors. 
:+1:

PR #619, opened 2015-10-15
no edX support for graphical slider

lamagnifica:
@explorerleslie , @mhoeber this fixes https://openedx.atlassian.net/browse/DOC-2291

### DIFF
--- a/en_us/olx/source/components/graphical_slider_tool.rst
+++ b/en_us/olx/source/components/graphical_slider_tool.rst
@@ -4,7 +4,7 @@
 Graphical Slider Tool
 *********************************************
 
-.. module:: xml_format_gst
+.. note:: EdX does not support this tool.
 
 
 Format description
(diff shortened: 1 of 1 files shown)


en_us/olx/source/components/graphical_slider_tool.rst
```diff
@@ -4,6 +4,8 @@
 Graphical Slider Tool
 *********************************************
 
+.. note:: EdX does not support this tool.
+
 .. module:: xml_format_gst
```
mhoeber: line 9 can be removed

mhoeber:
:+1:  thanks.  also, do we want this in table here:  http://edx.readthedocs.org/projects/edx-partner-course-staff/en/latest/exercises_tools/create_exercises_and_tools.html

lamagnifica:
well, @mhoeber that's an interesting question. All of the other tools in that table are also described in that guide. If we add this one to the table, shouldn't we also add it to the guide? I really don't understand how/why it ended up in the OLX guide only, and really it seems like we should be reducing the amount of attention paid to it rather than increasing its potential audience... Let me know if you want me to add it anywhere else, though.

mhoeber:
@lamagnifica  ok, makes sense to leave as is.

PR #635, opened 2015-10-22
Link fix for removed Google Drive Install

mhoeber:
@lamagnifica  quick fix for link to old Google drive install doc.


lamagnifica:
:+1:

PR #637, opened 2015-10-23
Unlimited signatories on certificate

mhoeber:
## DOC-2347

Minor change to allow unlimited signatories on certificate
### Date Needed:  10/26/15

Not to merge until SOL-1312 is in a release

Reviewers
- [ ] Subject matter expert: @mattdrayer 
- [x] Doc team review (dev edit): @lamagnifica 
### Testing
### Post-review
- [ ] Update change log
- [ ] Squash commits

### DIFF
--- a/en_us/shared/building_and_running_chapters/building_course/creating_course_certificates.rst
+++ b/en_us/shared/building_and_running_chapters/building_course/creating_course_certificates.rst
@@ -68,6 +68,7 @@ course.
 
 #. At the bottom of the page, select **Save Changes**.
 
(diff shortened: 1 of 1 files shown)


lamagnifica:
:+1: signatory gives me the heebie jeebies, but it looks like that ship has sailed...

mattdrayer:
:+1: 

Just as an FYI, there actually still is a limit of 100 on the number of signatories you can enter, because we felt we still needed a technical boundary even if practically speaking the constraint was unlikely to be encountered.

PR #659, opened 2015-10-30
Spell check, standards check, let the spacing changes happen.

lamagnifica:
@srpearce @catong 

is this super painful to review? should I do this for other guides? I figured I’d do the Install guide as a trial. It took about an hour to go through 30+ files, then review time…


en_us/install_operations/source/configuration/enable_certificates.rst
```diff
@@ -173,25 +168,26 @@ for eligible learners.
...
-     flag so that the certificate generation service contacts the lms on http instead of
-     https.
-
-3. View the certificate generation status for a course using ``gen_cert_report``. For example,
+
+   .. Note::
+     If the LMS is running on a server that does not have https support (such
+     as a locally run fullstack for testing) you will need to use the
+     ``--insecure`` flag so that the certificate generation service contacts
+     the lms on http instead of https.
```
catong: Add "on" in front of "https": "...contacts the lms on http instead of on https."

en_us/install_operations/source/mobile.rst
```diff
@@ -161,10 +164,10 @@ The procedures that follow assume that you have obtained an application ID,
 REST API key, and client key from Parse.
 
 ============================================
-Enable Push Notification on the Server  
+Enable Push Notification on the Server
```
catong, line 167: Should this be "Push Notifications" (plural)? It reads strangely otherwise.

catong:
@lamagnifica Not too difficult to review at all, esp. with the Split view. Thanks for doing this; I agree with all the changes, left a couple of small comments/questions. :+1:

en_us/install_operations/source/mobile.rst
lamagnifica, line 167: I agree, it seems awkward @catong. My recollection from convos with Lou is that the mobile feature is widely known as "push" or "push notification" (there are other types of notification as well). While it enables you to send (push) individual messages out to people who have your app installed, this doesn't enable the individual messages, but the entire feature... at least, that's how I tried to convince myself that this would read ok to the target audience of developers...

catong, line 167: @lamagnifica Yah, I can see your point. However, the configuration setting is actually named "PUSH_NOTIFICATIONS" so there is an argument for the plural :-)

PR #667, opened 2015-11-03
Remove unused certificate file

mhoeber:
@catong


catong:
If I am reading correctly, the file that is actually used in the Learner's Guide is SFD_certificates.rst, and replaces this file that we are deleting here.

catong:
:+1:

PR #675, opened 2015-11-08
fix description of problem_show event

ichuang:
problem_show is fired when the "show answer" button is clicked, not when a problem is shown.

See:  https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/js/src/capa/display.coffee#L336 and https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/js/src/capa/display.coffee#L38

PR #682, opened 2015-11-10
Ahodges/show problem

lamagnifica:
Replaces #675 contributed by @ichuang 

@catong or @srpearce sanity check please


srpearce:
:+1:

PR #696, opened 2015-11-23
Clean up two TBD references left lonely and abandoned

lamagnifica:
@srpearce @catong or @pdesjardins please sanity check these two rescues. 
Tests ran without error.


catong:
@lamagnifica LGTM :+1:

PR #717, opened 2015-12-03
Add newly discovered info about how translator comments can fail.

nedbat:


nedbat:
@lamagnifica @sarina OK?

sarina:
:+1:

en_us/developers/source/internationalization/i18n.rst
```diff
@@ -111,7 +111,24 @@ instead::
     message = _("Welcome!")
 
 "Translators" comments will work in these places too, so don't be shy about
-providing clarifying comments to the translators.
+providing clarifying comments to the translators. When writing translators
```
lamagnifica, line 114: would it make sense to consistently icap the Translators in the phrase Translators comments ?

lamagnifica:
:+1: please add a change log entry.
Also, just to raise awareness _for the future_, the team has developed style guidelines to make translation go more smoothly. Your addition has a consistent style with the rest of this file as it is now, so no worries for this PR, but going forward we are trying to minimize use of contractions, compound sentences (such as those with two clauses separated by a colon) and use of sentence fragments that end with a colon.

PR #728, opened 2015-12-09
Make Travis fail builds on warnings.

nedbat:
I also fixed the two existing image reference warnings.

### DIFF
--- a/en_us/shared/subsections/open_response_assessments/OpenResponseAssessments.rst
+++ b/en_us/shared/subsections/open_response_assessments/OpenResponseAssessments.rst
@@ -283,7 +283,7 @@ Rubrics consist of *criteria* and *options*.
 
 * Each option has a *name*, an *explanation*, and a *point value*.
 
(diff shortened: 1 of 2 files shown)


nedbat:
@lamagnifica @catong @srpearce @pdesjardins This changes Travis to fail (red X) if there are any warnings or errors at all.  Previously, Travis would give you a red X on errors, but warnings were allowed.
@clytwynec in case I did this sub-optimally, advice welcome.

catong:
@nedbat the image paths fix look good; thanks for taking care of them! :+1: for that; I can't comment on the test fix, except that I agree with the wording change.

nedbat:
@catong (et al.) On the script itself, I need your OK with the process change, since you will be living with the results of this more than I will.

catong:
@nedbat The process change is fine with me; I agree that we need the Travis test to show as Failed not just when there are errors, but also when there are warnings. Thanks!

lamagnifica:
local tests run without error after these changes. 
:+1:

PR #735, opened 2015-12-11
Documentation for Badges v2 (DOC-2486, DOC-2496)

catong:
This PR adds configuration documentation for v2 of the badging feature.
Tracked in https://openedx.atlassian.net/browse/DOC-2486, DOC-2496

Technical reviewers: please review for factual accuracy, especially where there might have been changes between specs and final implementation, changes from v1 feature. Thanks!

Draft documentation output is here: http://draft-install-guide-badging.readthedocs.org/en/latest/configuration/enable_badging.html
### Reviewers
- [x] Subject matter expert: @Kelketek 
- [ ] Subject matter expert (optional): @antoviaque 
- [x] Product owner: @pbaruah 
- [x] Doc team review: @lamagnifica 

FYI @stroilova 
### Testing
- [x] Run ./run_tests.sh without warnings or errors
### Post-review
- [x] Squash commits


en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +84,66 @@ Server.
...
+
++++++++++++++++++++++++++++++++++++++++++++++
 Specify a Badge Issuer for Your Organization
-*****************************************************************
++++++++++++++++++++++++++++++++++++++++++++++
 
-Log in to your installation of Badgr Server and add an issuer of Open Badges
-for your organization.
+.. note:: This step is not required if you do not use Badgr Server for your
+   badging backend.
```
catong: @Kelketek Is this true? Not required unless using Badgr Server?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +84,66 @@ Server.
...
+   well as for all checks leading up to the awarding of the badge, including
+   creation of the badge specification on the target if it does not yet exist.
+   The ``award`` function should either return a BadgeAssertion object or raise
+   an exception if the award cannot be given. By default, a base backend object
+   called ``BadgeBackend`` exists, and contains a dummy version of the ``award``
+   function.
+
+#. In the BadgeAssertion object, use the ``backend`` value to reference the
+   badging backend that is used to generate the badge.
+
```
catong, line 202: @Kelketek Can you pls provide an example that I could include?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +84,66 @@ Server.
...
++++++++++++++++++++++++++++++++++++++++++++++++++
+Specify a Badging Backend Other Than Badgr Server
++++++++++++++++++++++++++++++++++++++++++++++++++
+
+By default, Open edX uses Badgr Server as the badging backend. To use Badgr
+Server, see :ref:`Install Badgr Server` and :ref:`Specify a Badge Issuer`.
+
+To specify a different badging backend, follow these steps.
+
+#. In settings.py, specify BADGING_BACKEND. This value is a .-separated python
```
catong: @Kelketek what is the full path to settings.py?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,9 +188,78 @@ which are located one level above the ``edx-platform`` directory.
...
+For details, see :ref:`Create Course Event Badges`.
+
+
+.. _Create Course Completion Badges:
+
+**********************************************************
+Create Course Completion Badges for Your Open edX Instance
+**********************************************************
+
+.. Is it still true that we ship open edx with default images?
```
catong: @Kelketek Any change to this statement that we ship with default images?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +84,66 @@ Server.
...
+   an exception if the award cannot be given. By default, a base backend object
+   called ``BadgeBackend`` exists, and contains a dummy version of the ``award``
+   function.
+
+#. In the BadgeAssertion object, use the ``backend`` value to reference the
+   badging backend that is used to generate the badge.
+
+
+.. _Enable Badges in Studio and LMS:
+
```
catong, line 205: @Kelketek 
In step 1 below (line 154 not in this PR), we say "set the value of `ENABLE_OPENBADGES` to `True`. Does this need to be modified if Badgr Server is not the badging backend? Any other impacts of not using Badgr Server?

Kelketek, line 205: @catong No. The badging backend setting would not be used if this isn't enabled.

en_us/install_operations/source/configuration/enable_badging.rst
Kelketek, line 202: I'm not quite sure what to give you for an example. But here in the code is what is being referred to:

https://github.com/edx/edx-platform/blob/fecaad58829032341c2e14ce372fcf3fec671092/lms/djangoapps/badges/models.py#L91

This instantiates the backend from a BadgeClass, for convenience.

en_us/install_operations/source/configuration/enable_badging.rst
Kelketek: @catong See my comments elsewhere-- settings.py is the general name for the Django settings file-- we split ours into common.py, aws.py, and the _.env/_auth.json files.

en_us/install_operations/source/configuration/enable_badging.rst
Kelketek: Correct-- though other badging backends might have their own issuer setup.

en_us/install_operations/source/configuration/enable_badging.rst
Kelketek: @catong No, but I heard Devops was having trouble with the migration that creates them. :)

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,9 +188,78 @@ which are located one level above the ``edx-platform`` directory.
...
+*****************************************
+Enable Badges Within Each Course
+*****************************************
+
+The ability to issue course completion badges is enabled by default for all
+courses, but can be turned off or on again using an advanced setting in
+Studio. For details, see :ref:`opencoursestaff:Enable Badges in Course` in
+*Building and Running an Open edX Course*.
+
+.. note:: The course-level setting to enable badges does not affect badges
```
Kelketek: XBlock-level badges also would still be awarded in the current implementation. This may need to be changed. @antoviaque ?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -120,34 +272,171 @@ Configure Badges and Badge Images for Your Open edX Instance
...
+
+#. Access the Django Administration website for your instance of Open edX. To
+   do this, go to ``https://<host name of your Open edX instance>/admin``. For
+   example, this might be ``https://YourOrganization.org/admin``
+
+#. Under **Site Administration** > **Badges**, select **Badge Classes**.
+
+#. Add a badge class for each course event for which you want to issue badges.
+   An example of a course event associated with a badge is "Badge for
+   enrolling in 5 courses".
```
Kelketek: This phrasing is a bit awkward-- the quote marks make it sound like they're expected to paste this string in somewhere.

catong: Revised.

en_us/install_operations/source/configuration/enable_badging.rst
catong, line 205: @Kelketek I should rephrase my question so that I understand the answer :-) If Badger Server is not the badging backend, should the value of `ENABLE_OPENBADGES` be `False`?

catong:
@Kelketek Ready for your follow-up review. Replaced "backend" with "badge generator" and updated the default issuing component name. Also a few specific questions inline for you. Thanks for your help!

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +97,99 @@ Server.
...
+By default, Open edX uses Badgr Server as the badge generator. To use Badgr
+Server, see :ref:`Install Badgr Server` and :ref:`Specify a Badge Issuer`.
+
+To specify a different badge generator, follow these steps.
+
+#. In the ``lms.env.json`` and ``cms.env.json`` files, set the value of
+   ``BADGING_BACKEND`` as a dot-separated python path specification to the
+   object that you use to create and assign badges.
+
+#. Add the BadgeBackend object as a subclass to
```
catong: @Kelketek Should "BadgeBackend object" be "BadgingBackend object"?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +97,99 @@ Server.
...
+   ``lms/djangoapps/badges/backends/base.py``.
+
+   A BadgeBackend object must include the ``award`` function.
```
catong: Should "BadgeBackend object" be "BadgingBackend object"?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +97,99 @@ Server.
...
+        """
+        user = models.ForeignKey(User)
+        badge_class = models.ForeignKey(BadgeClass)
+        data = JSONField()
+        backend = models.CharField(max_length=50)
+        image_url = models.URLField()
+        assertion_url = models.URLField(
+
+#. When you have finished modifying your configuration files for the badge
+   generator, run database migrations.
```
catong: @Kelketek is it correct/necessary to specify the run database migrations and restart Studio and LMS steps here?

en_us/install_operations/source/configuration/enable_badging.rst
catong: @Kelketek I don't understand the implications of your question. Will wait for @antoviaque's answer.

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +234,254 @@ which are located one level above the ``edx-platform`` directory.
...
+   * Mode: The course mode for the course associated with this badge, if
+     applicable.
+
+   * Image: The badge image to use for this badge. Badge images should be
+     square .png files less than 250KB in size.
+
+   An example of a badge class configuration might have the following values.
+
+   * ``slug``: ``enrolled_three``
+   * ``issuing_component``: ``openedx__course``
```
catong, line 440: I added ``issuing_component" here

en_us/install_operations/source/configuration/enable_badging.rst
Kelketek: @catong Yes, since BadgeBackend is the name of the superclass in the file you're listing there.

The order of the items in this list may need to be reversed-- you'll probably want to create the badge backend class before pointing your settings to it :)

Kelketek: Or rather, the first item in the list should be moved to the end-- it need not be fully reversed.

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +97,99 @@ Server.
...
+        """
+        __metaclass__ = ABCMeta
+
+        @abstractmethod
+        def award(self, badge_class, user, evidence_url=None):
+        """
+        Create a badge assertion for the user using this badge generator.
+        """
+
+#. In the BadgeAssertion object, use the ``backend`` value to reference the
```
Kelketek: This step isn't very clear to me-- it's better to say that when generating the BadgeAssertion object in the award step, to make sure they set the backend field to be the name of the backend class they made.

en_us/install_operations/source/configuration/enable_badging.rst
Kelketek: @catong Restarting the services is good. There shouldn't be a need to run migrations just for changing the backend, though.

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +234,254 @@ which are located one level above the ``edx-platform`` directory.
...
+     component, you might define the issuing component as
+     ``platform__course``; if badges are issued based on activity in course
+     discussions, you might define the issuing component as
+     ``platform__discussions``.
+
+   * Display name: The human readable badge name that is used when badges are
+     shown to learners, for example, in the Accomplishments view of learners'
+     profiles.
+
+   * Course ID: The course identifier for the course associated with this
```
Kelketek: This will need to be blank for the included events.

Kelketek:
@catong Just a few notes for you. Let me know when you've got them addressed and I'll take another look.

catong:
@Kelketek and @pbaruah I add sections relating to creating new badges. Please take a look. Also FYI I'm going to keep the scope of this PR to configuration documentation only, even though initially I made some changes to the "Building and Running a course" guide. For further changes to the course team documentation I will start a new PR.

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -4,44 +4,91 @@
...
+for a learner. Make sure certificates are enabled on your Open edX instance.
+For details, see :ref:`Enable Certificates`.
+
+.. _Specify a Badge Generator:
+
+=========================
+Specify a Badge Generator
+=========================
+
+By default, Open edX uses Badgr Server as the badging generator. If you do not
```
Kelketek: In the title you use "Badge Generator", but in the paragraph you use "Badging Generator". "Badge Generator" is probably the better term to use.

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,458 @@ which are located one level above the ``edx-platform`` directory.
...
+
+   * Issuing component: For the three course event badges that are provided
+     with Open edX, the value for issuing component must be ``openedx__course``.
+     For new custom badge types that you create, specify an issuing component
+     value that identifies the software component responsible for issuing the
+     badge. For example, if badges are issued by the course management
+     component, you might define the issuing component as
+     ``platform__course``; if badges are issued based on activity in course
+     discussions, you might define the issuing component as
+     ``platform__discussions``.
```
Kelketek, line 419: It may be worth noting that this field has the same constraints as slug-- numbers, lowercase alphabetical letters, underscores, and hyphens.

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,458 @@ which are located one level above the ``edx-platform`` directory.
...
+Creating New Badges for Your Open edX Instance
+**********************************************
+
+You can create new badges that are generated when a particular XBlock-related
+or course-related action occurs.
+
+Before you create new badges, you should understand the following concepts.
+
+* Badge Class - The specification of the badge that is to be awarded. Each
+  badge class should include parameters. as described in Step 4 in
```
Kelketek: "Each badge class should include parameters. as described in Step 4 in"

This wording is a confusing, and there's a full stop in the middle of a sentence.

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,458 @@ which are located one level above the ``edx-platform`` directory.
...
+  For the customizable :ref:`course event badges<Course Event Badges>` that
+  are included with the Open edX platform, the value for ``issuing_component``
+  must be ``openedx__course``. For :ref:`course completion badges<Course
+  Completion Badges>` that are included in the Open edX platform, the issuing
+  component value should be empty.
+
+  The combination of a badge slug and issuing component uniquely identifies an
+  awarded badge.
+
+* Award method - This function is responsible for awarding a badge to a
```
Kelketek: This description of the award method is more apt for describing the implementation for a badge generator. The award method on a badge class should return a Badge Assertion object or raise an exception-- this much is true, and it does call the Badge Generator's award method. But information about the default backend is not especially relevant here-- that's an abstraction layer away, and not especially relevant to a developer looking to create a new type of badge.

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,458 @@ which are located one level above the ``edx-platform`` directory.
...
+
+To create new badges for use within your Open edX platform, follow these steps.
+
+#. Create a Badge Class specifying the details of the badge you want to award.
+   You can do this using the ``get_badge_class`` method from the ``badges``
+   app, or create a badge class in Django Admin. For Django Admin
+   instructions, see steps 1-5 in :ref:`Create Course Event Badges`.
+
+   ``BadgeClass.get_badge_class`` creates the requested badge class if one does not
+   already exist that has the same combination of ``slug`` and ``issuing_component``.
```
Kelketek: 'slug', 'issuing_component' and 'course_id' are unique together. That allows someone to award a specific XBlock badge for each course the XBlock is in, OR to award it if they manage to complete the XBlock in any course (should no course ID be used as a qualifier).

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,458 @@ which are located one level above the ``edx-platform`` directory.
...
+   Badge classes are uniquely identified by a combination of their ``slug`` and
+   ``issuing_component`` fields.
+
+   If a badge class already exists with the same combination of ``slug`` and
+   ``issuing_component`` that is in the request, the existing badge is returned.
+   No new badge class is created, and no changes are made to the values of the
+   existing badge class.
+
+#. In your badge generator, make sure the ``BadgingBackend`` object includes
```
Kelketek: This part of the documentation probably shouldn't involve anything about the BadgingBackend. An XBlock developer, for instance, shouldn't need to know what backend the particular OpenedX instance is using. If the BadgingBackend object doesn't have an award function, that's beyond the duty of someone making a particular badge class.

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,458 @@ which are located one level above the ``edx-platform`` directory.
...
+     - Endpoint
+   * - Get the list of assertions for a specified user
+     - GET
+     - /api/badges/v1/assertions/user/{username}
+   * - Get the list of assertions for a specified user, filtered by badge class
+     - GET
+     - /api/badges/v1/assertions/user/{username}/{slug}/{issuing_component}
+   * - Get the list of assertions for a specified user, filtered by course id
+     - GET
+     - /api/badges/v1/assertions/user/{username}/{courseid}
```
Kelketek: we've changed how this works. There's just one URL endpoint for all of these, with different query parameters.

http://pr10556.sandbox.opencraft.com/api/badges/v1/assertions/user/honor/ shows the current API documentation built in. Note the listing of params. An example query might be:

http://pr10556.sandbox.opencraft.com/api/badges/v1/assertions/user/honor/?issuing_component=edx__course&slug=robin

Which looks for the issuing_component `edx__course` and the slug `robin`.

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,458 @@ which are located one level above the ``edx-platform`` directory.
...
+
+.. list-table::
+   :widths: 40 60
+   :header-rows: 1
+
+   * - Parameters
+     - Description
+   * - slug
+     - (Optional) If used, filters by the badge class identified by this slug. Unless ``issuing component`` is also specified, assumes a null/empty issuing component.
```
Kelketek: Is this section of the documentation for the Badge Class endpoint? The URL is not provided.

en_us/install_operations/source/configuration/enable_badging.rst
catong: Oops. Fixed throughout, thanks.

en_us/install_operations/source/configuration/enable_badging.rst
catong: @Kelketek Thanks. So no need to mention the `award` function in these steps at all?

Kelketek: There are two award functions, one on the badge class and one on the backend. The one on the badge class instantiates the backend and calls the backend's award function so that the developer doesn't have to interact with the backend directly. The award method on a badge class just takes the user and, optionally, the evidence_url as arguments.

Kelketek: That is to say, the award method is still relevant here for awarding the badge to the intended user. I don't know if its mention is best placed here or elsewhere.

catong: What action might a developer need to do for making sure the award method on the badge class is going to work? That's probably what I should put here, right?

Kelketek: @catong If the award method is called correctly but doesn't work, the developer's code should probably crash or just ignore the failure. It would indicate something's wrong under the hood, and that something is not the responsibility of the badge developer-- it would be the responsibility of whoever is configuring the instance (in all likelyhood. It is possible that the developer made a mistake but if they're all the way here their information about the badge class should have validated. If they hand the award method bad data, it will crash and give a sane traceback, and it should do so, but I don't think that's what you're talking about here).

The biggest reason why the award method might fail would probably be the Badgr Server being down, or the API key not being set correctly-- both of which are ops issues.

catong: @Kelketek do they provide the evidence_URL?

Kelketek: @catong Yes, though in most cases I imagine the evidence_url wouldn't be included, since there's not really a publicly facing interface specific XBlocks present to show a user did something to the outside world. In the case of course completions, we use the web certificate URL as the evidence URL. It's not explicitly required by the OpenBadges Spec, hence it being an optional argument.

catong:
@Kelketek I've made edits per your comment. Please take a further look when you have a chance. FYI @pbaruah Probably a good time also for you to have a look. The draft HTML link in the PR template comment is updated to reflect the latest updates.

en_us/install_operations/source/configuration/enable_badging.rst
antoviaque: @Kelketek If badges are disabled for a given course, it would be more consistent to ensure the badges generated by XBlocks and scoped to that course would be disabled by that setting, yes - imho if it's not the case currently, it should be changed.

CC @pbaruah

Kelketek:
@catong One minor fixup comment, but otherwise :+1:

pbaruah:
@catong - this looks great. Are you planning to mention that once badges are generated, they will appear in the user profile, accomplishments tab?

:+1:

catong:
@pbaruah I'd like to keep mention of those other areas separate because they probably will release at different times. Will add to documentation when they are definitely available.

catong:
@lamagnifica or @srpearce @pdesjardins whoever has time to do a doc review, please?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -4,44 +4,91 @@
...
+In addition to badges that are awarded for completing single courses, you can
+also issue badges for various cross-course events. For example, you can create
+badges to be awarded when any of the following events occur.
+
+* A learner enrolls in a certain number of courses.
+* A learner receives a completion certificate for a certain number of courses.
+* A learner receives a completion certificate for every course in a specified
+  list of courses.
+
+For more details, see :ref:`Course Event Badges`.
```
lamagnifica: details --> information about badges for cross-course events

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -4,44 +4,91 @@
...
 
 Open edX supports Open Badges, an open standard originally developed by the
 Mozilla Foundation. For more information about Open Badges, see
-http://openbadges.org/.
+http://openbadges.org/. For details about using a badging service other than
```
lamagnifica: details --> more information

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -4,44 +4,91 @@
...
+Open Badges, see :ref:`Specify a Badge Generator Other Than Badgr Server`.
 
-Enabling the badges feature on your instance of Open edX involves the
-following set up and configuration tasks.
+For details about creating custom badges, see :ref:`Create New Badges`.
```
lamagnifica: details again

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -4,44 +4,91 @@
...
+.. _Make Sure Certificates are Enabled:
+
+=======================================
+Make Sure Certificates Are Enabled
+=======================================
+
+Badge generation depends on certificate generation. Badges for course
+completion are automatically generated when a course certificate is generated
+for a learner. Make sure certificates are enabled on your Open edX instance.
+For details, see :ref:`Enable Certificates`.
```
lamagnifica: global search and replace on details?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +98,97 @@ Server.
    address, to allow the Open edX LMS and services such as Mozilla Backpack to
    contact Badgr Server.
 
-*****************************************************************
+If you do not use Badgr Server, you can configure a different badge generator,
+see :ref:`Specify a Badge Generator Other Than Badgr Server`.
```
lamagnifica: break into two sentences?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +98,97 @@ Server.
...
+.. _Specify a Badge Issuer:
+
++++++++++++++++++++++++++++++++++++++++++++++
 Specify a Badge Issuer for Your Organization
-*****************************************************************
++++++++++++++++++++++++++++++++++++++++++++++
 
-Log in to your installation of Badgr Server and add an issuer of Open Badges
-for your organization.
+.. note:: This step is not required if you do not use Badgr Server for your
```
lamagnifica: double negative makes this hard to understand. also, from what is written above, it sounds as though the second "not" in the sentence might be an error? 
or maybe it's the opposite? This step is only required if you use Badgr Server (?)

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +98,97 @@ Server.
...
+   badge generator.
+
+If you are using Badgr Server, log in to your installation of Badgr
+Server and add an issuer of Open Badges for your organization.
 
 For more information about issuing Open Badges, see the `Issuing Badges`_ topic
 on the Mozilla wiki.
```
lamagnifica, line 119: because of references like this, I'd definitely suggest adding (near top of topic) something like 

.. important::
 The tasks described by this topic rely on the use of third-party services and
 software. Because the services and software are subject to change by their
 owners, the information provided here is intended as a guideline and not as
 exact procedures.

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +98,97 @@ Server.
...
+   ``lms/djangoapps/badges/backends/base.py``.
+
+   A BadgingBackend object must include the ``award`` function.
+
+   ``award(badge_class, user_id, evidence_url=None)``
+
+   where ``badge_class`` is the definition of the badge class for which the
+   BadgeAssertion is created, ``user_id`` is an ID for the user that the
+   BadgeAssertion is to be created for, and ``evidence_url`` is an optional URL
+   for the location where the badge evidence can be viewed.
```
lamagnifica: could this be bulleted out, with each line starting with the variable followed by the definition?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +98,97 @@ Server.
...
+
+   This function is responsible for the server awarding a badge to a learner,
+   as well as for all checks leading up to the awarding of the badge,
+   including creation of the badge specification on the target if it does not
+   yet exist. The ``award`` function should either return a BadgeAssertion
```
lamagnifica: break into two sentences?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -51,16 +98,97 @@ Server.
...
+        """
+
+#. In the BadgeAssertion object that is generated by the ``award`` function, make
+   sure the ``backend`` value is the name of the BadgingBackend subclass.
+
+   .. code-block:: bash
+
+    class BadgeAssertion(models.Model):
+        """
+        Tracks badges on our side of the badge baking transaction
```
lamagnifica, line 186: baking? that's hilarious

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -85,9 +213,11 @@ which are located one level above the ``edx-platform`` directory.
...
      Badgr Server must be installed at a publicly accessible IP address.
 
-   * ``BADGR_ISSUER_SLUG``: A string that is the slug for the Badgr issuer. The
-     slug can be obtained from the URL of the Badgr Server page that displays
-     the issuer. For example, in the URL ``http://exampleserver.com/issuer/test-issuer``, the issuer slug is ``test-issuer``.
+   * ``BADGR_ISSUER_SLUG``: A string that is the slug for the Badgr issuer.
+     The slug can be obtained from the URL of the Badgr Server page that
+     displays the issuer. For example, in the URL
+     ``http://exampleserver.com/issuer/test-issuer``, the issuer slug is
+     ``test- issuer``.
```
lamagnifica: extra space?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+Enable Badges Within Each Course
+*****************************************
+
+The ability to issue course completion badges is enabled by default for all
+courses, but can be turned off or on again using an advanced setting in
+Studio. For details, see :ref:`opencoursestaff:Enable Badges in Course` in
+*Building and Running an Open edX Course*.
+
+.. note:: The course-level setting to enable badges does not affect badges
+   that are issued for completing multiple courses or enrolling in multiple
```
lamagnifica: maybe delete the first "multiple courses"?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+   courses.
+
+
+.. _Badge Types:
+
+*******************************************
+Configure Badges for Your Open edX Instance
+*******************************************
+
+You can configure several types of badges for issuing to learners in your Open
```
lamagnifica: for issuing --> to issue

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+details, see :ref:`About Creating New Badges`.
+
+.. _Course Completion Badges:
+
+========================
+Course Completion Badges
+========================
+
+For courses that have badges enabled, learners receive a course completion
+badge at the same time as they receive a course certificate, and have the
```
lamagnifica, line 291: as they --> that they ?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+option of sharing their badges to a badging site such as Mozilla Backpack.
+
+You can configure a different course completion badge for each course mode
+that you support (for example, "professional" or "audit"), including
+specifying a different badge image. For details, see :ref:`Enable Badges For a
```
lamagnifica: break into two sentences? ...you support, such as professional or audit badges. For example, you can specify a different image for each of the badges you configure.

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+Course` and :ref:`Create Course Completion Badges`.
+
+
+.. _Course Event Badges:
+
+========================
+Course Event Badges
+========================
+
+Course event related badges are awarded across courses. Customizable badges are
```
lamagnifica: strike "related"?
are course event badges the same as customizable badges? why the terminology switch?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+
+**********************************************************
+Create Course Completion Badges for Your Open edX Instance
+**********************************************************
+
+.. important:: Default images are supplied for course completion badges. You
+   must replace the default images with your organization's own badge images
+   before any badges are issued. When the first badge is issued for a given
+   course, badge images are uploaded to Badgr Server. All badges issued in
+   future for this course will use the original badge image, even if you
```
lamagnifica: the original --> this original (?)

not quite understanding at why you _must_ change the default image -- maybe show what the default is, and say, hey, you REALLY don't want all your learners to get this, do you?

maybe more active construction here would help too, to underscore that badge issuing is automated and you can't stop the ship from sailing?? (if that's correct)

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+   configuration.
 
 #. Access the Django Administration website for your instance of Open edX. To
    do this, go to ``https://<host name of your Open edX instance>/admin``. For
    example, this might be ``https://YourOrganization.org/admin``.
 
-#. Under **Site Administration** > **Certificates**, define a Badge Image
-   Configuration for each course mode on your platform for which you want to
-   issue badges. For example, "honor" and "verified".
+#. Under **Site Administration** > **Badges** > **Course complete image
```
lamagnifica: "Under" construction is unusual for us...  maybe , Select ... , and then define...
(though I always love the menu > option > setting construction and would like us to move toward using it more!)

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+   configurations**, define a course complete image configuration for each
+   course mode on your platform for which you want to issue badges upon course
+   completion. Examples of course modes are "professional" or "audit track".
```
lamagnifica: move track out of the quotes to match example above?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
       organization's own badge images before any badges are issued.
 
-#. Optionally, define a default image for any course modes that do not have an
-   explicitly specified badge image. Select **Default** in the badge image
-   configuration.
+#. Optionally, you can define a badge image that will be used as the default
+   badge image for any course modes without an explicitly specified badge
+   image.
+
+   In the course complete image configuration that references the image you
```
lamagnifica: To do so, in...

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+   want to use as a default, select the **Default** checkbox. After you save
+   the configuration, this badge image is used for any course completion badge
+   configurations that do not have a badge image explicitly specified.
 
    .. note:: You can specify only one default badge image.
 
 #. Save each configuration parameter and exit the Django Administration
    website.
```
lamagnifica, line 364: does this upload the specified "icon" images? maybe state that explicitly, and where they get stored if you want to review/confirm them?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+
+Open edX provides several customizable course event badges that can be awarded
+when any of the following events occur.
+
+* A learner enrolls in a certain number of courses.
+* A learner receives a completion certificate for a certain number of courses.
+* A learner receives a completion certificate for every course in a specified
+  list of courses.
+
+These course event badges can be customized with your parameters and badge
```
lamagnifica: yeah, the "can be" vs. "must" construction around changing images and making customizations could be clearer

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+images. To customize any of the course event badges, follow these steps.
+
+#. Access the Django Administration website for your instance of Open edX. To
+   do this, go to ``https://<host name of your Open edX instance>/admin``. For
+   example, this might be ``https://YourOrganization.org/admin``
+
+#. Under **Site Administration** > **Badges**, select **Badge Classes**.
```
lamagnifica: I'd just go, "Select ... > ... > ... ."

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+
+#. Add a badge class for each course event for which you want to issue badges.
+   Examples of course events might be enrolling in five courses, or completing
+   three required courses.
+
+#. For each badge class, set the following parameters.
+
+   * Slug: A unique identifier that you choose to identify the badge class.
+     This identifier can contain only numbers, lowercase alphabetical letters,
```
lamagnifica: strike alphabetical (redundant with letters)?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+     underscores, or hyphens. The slug, combined with the Issuing Component value,
```
lamagnifica: inconsistent capitalization of Issuing Component here to below; if this is in the UI maybe keep icaps but make bold?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+     uniquely identifies a badge.
+
+   * Issuing component: Identifies the part of the software that is issuing the
```
lamagnifica: software --> platform (to make consistent with your examples below, and tie more closely to the edX product?)

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+#. When you have finished defining the badge class, select **Save**.
+
+#. Next, you create a new course event badge configuration that defines all of
+   the course event badges that you want to issue.
+
+   .. note:: You can create more than one course event badge configuration,
+      but only the most recently activated course event badge configuration
+      is used.
+
+#. Under **Site Administration** > **Badges**, select **Course event badges
```
lamagnifica: throughout
you're carving out new style territory here! it's awesome!

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+
+   * Courses completed: Define badges to be awarded for completing a certain
+     number of courses, or completion of specific courses. Define one badge
+     per line. On each line, enter the number of courses that must be completed,
+     followed by a comma and then the slug of the badge class to associate
+     with this badge.
+
+     For example, to configure two badges, one that is awarded when a learner
+     completes 3 courses, and another that is awarded when a learner completes
+     8 courses, you would add two lines to the **Courses completed** field.
```
lamagnifica: strike "would"

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+
+     ``3,completed_three``
+     ``8,completed_eight``
+
```
lamagnifica, line 473: how does this look for the "completion of specific courses" use case mentioned on line 438?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+     you previously defined in badge classes.
+
+   * Courses enrolled: Define badges to be awarded for enrolling in a certain
+     number of courses, or enrolling in specific courses. Define one badge per
+     line. On each line, enter the number of courses that must be enrolled in,
+     followed by a comma and then the slug of the badge class to associate
+     with this badge.
+
+     For example, to configure a badge that is awarded when a learner enrolls
+     in 5 courses, you would add this definition.
```
lamagnifica: strike would

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+
+     ``5,enrolled_five``
```
lamagnifica, line 486: again, an example of the other use case would be helpful

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+   * Course groups: Define badges to be awarded for completing a list of
+     specific courses. Define one badge per line. On each line, enter the slug
+     of the badge class, a comma, then the list of course keys.
+
+     For example, to configure a badge that is awarded when a learner
+     completes the 3 prerequisite courses in a series, you would add this
+     definition.
+
+     ``prereq_computerscience_badge_slug,course1_identifier,course2_identifier,
+     course_3_identifier``
```
lamagnifica: line break ok here?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+
+     where ``prereq_computerscience_badge_slug`` is a badge slug that you
+     previously defined in a badge class, and ``course1_identifier``,
+     ``course2_identifier``, and ``course3_identifier`` are the Course IDs for
+     the three courses that must be completed for this badge.
+
+#. When you have finished defining badges in the configuration, select **Save**.
+
+#. To activate this configuration, select **Enabled** at the top left of the
+   configuration page.
```
lamagnifica, line 508: careful of directing to a location per Mark's a11y guidance

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+
+.. important:: Only one course event badge configuration can be active at any
+   one time. The most recently updated active configuration is used.
```
lamagnifica: wait, what? it sure sounds like I can set up a whole bunch of different situations that will result in learners getting badges up around line 354... this important seems like it's in the wrong place, and maybe like a "course event badge configuration" needs to be more clearly defined!

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+* Slug - This field in the BadgeClass model uniquely identifies the badge
+  class.
+
+* Issuing Component - This field in the BadgeClass model identifies the part
+  of the software that is issuing the badge. For example, the name of the
+  XBlock where the occurrence of some event triggers the awarding of a badge.
+
+  For the customizable :ref:`course event badges<Course Event Badges>` that
+  are included with the Open edX platform, the value for ``issuing_component``
+  must be ``openedx__course``. For :ref:`course completion badges<Course
```
lamagnifica: looks like two underscores? if so, maybe explicitly state that?

also, what? above, you give examples of what this value should look like, such as platform_discussions
Maybe I'm just not up to speed on the difference between what they were doing in the last procedure (Create Course Event Badges for Your Open edX Instance) and what they're describing here (Creating New Badges for Your Open edX Instance), and how those two things are different -- put that more up front some how? (like add on line 499. "The edX platform does not supply defaults for these badges, but you can configure them to provide a customized badging experience for your learners" or some less jargony thing?) and up on line 366-ish put something like, you can also set up customized badges to award for other events that a learner completes, go see that other topic down there)

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+============================================
+
+To create new badges for use within your Open edX platform, follow these steps.
+
+#. Create a Badge Class specifying the details of the badge you want to award.
+   You can do this using the ``get_badge_class`` method from the ``badges``
+   app, or create a badge class in Django Admin. For Django Admin
+   instructions, see steps 1-5 in :ref:`Create Course Event Badges`.
+
+   ``BadgeClass.get_badge_class`` creates the requested badge class if one does
```
lamagnifica, line 585: is this a step?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+  For details and additional parameters, see :ref:`Badges API Endpoints`.
+
+
+.. _Badges API Endpoints:
+
+=====================
+Badges API Endpoints
+=====================
+
+The Badges API supports the endpoint ``GET /api/badges/v1/assertions/user/{username}/``, with which you can use the following query parameters.
```
lamagnifica: grr wrapping
Maybe two sentences?
The Badges API supports the `GET /api/badges/v1/assertions/user/{username}/`  endpoint. You can use the following query parameters with this endpoint.

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+  return a list of all assertions that the specified user has received for the
+  badge class.
+
+* Use the Badges API GET method to return a list of assertions for a particular
+  username. For example, use ``GET /api/badges/v1/assertions/user/{username}/``.
+  For details and additional parameters, see :ref:`Badges API Endpoints`.
+
+
+.. _Badges API Endpoints:
+
```
lamagnifica, line 667: ? why put API descriptions here? is it our API? or another company's API? I'd xref to an API reference doc either way...

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+=====================
+Badges API Endpoints
+=====================
```
lamagnifica: maybe singular, something more like "Supported Badges API Endpoint" (?)

en_us/install_operations/source/configuration/enable_badging.rst
lamagnifica: maybe state here that all of the parameters are optional, rather than repeating 3x?

en_us/install_operations/source/configuration/enable_badging.rst
```diff
@@ -105,49 +235,442 @@ which are located one level above the ``edx-platform`` directory.
...
+
+   * - Parameters
+     - Description
+   * - slug
+     - (Optional) If used, filters by the badge class identified by this slug. Unless ``issuing component`` is also specified, assumes a null/empty issuing component.
+   * - issuing component
+     - (Optional) If used, also requires ``slug`` to be specified. Filters by
+       the badge class.
+   * - courseid
+     - (Optional) If used returns only badge assertions that were awarded as
```
lamagnifica: If used,

en_us/shared/set_up_course/creating_course_certificates.rst
```diff
@@ -331,33 +336,33 @@ manually remove unused images. For information, see
 
 .. only:: Open_edX
```
lamagnifica, line 337: not sure how the spacing works out, is this ^ still needed?

en_us/shared/set_up_course/creating_course_certificates.rst
```diff
@@ -331,33 +336,33 @@ manually remove unused images. For information, see
...
-  #. Locate the **Issue Open Badges** policy key. The default value is
-     ``True``.
+ #. Locate the **Issue Open Badges** policy key. The default value is ``True``.
 
-  #. Change the setting to ``False`` and save your changes.
+ #. Change the setting to ``False`` and save your changes.
 
-  To enable badging for your course if it has previously been disabled, change
-  the value of the key to ``True``.
+ To enable badging for your course if it has previously been disabled, change
```
lamagnifica: has ... been --> was

en_us/shared/set_up_course/creating_course_certificates.rst
lamagnifica, line 337: or maybe there is a Open_edX only note about working with your system admin to make sure that the design you want for your badge is set up well in advance of any learner being awarded a certificate?
(similarly for partners, work with your PM?)

lamagnifica:
That's a big, complex topic. I'm not sure I understand the nuances yet. As always, use edits that make sense! :+1:

en_us/install_operations/source/configuration/enable_badging.rst
catong: Tried to clarify a bit by rewording. Open edX ships with default badge images but they are branded for open edx. Presumably each open edx instance would want badges that are designed for their organization/institution. When the first badge is issued for a course, the badging server gets the image that's specified at that moment. Even if the badge image is changed afterwards, all future badges of that type still use the old image.

en_us/install_operations/source/configuration/enable_badging.rst
catong: In this case "must" is wrong; it's only the course completion badges that we provide default images for. Hm, I'll still try to reword to lessen confusion.

en_us/install_operations/source/configuration/enable_badging.rst
catong, line 473: This case is covered by "Course Groups" at line 467.

en_us/install_operations/source/configuration/enable_badging.rst
catong, line 486: This configuration only allows defning the number of courses enrolled in. You use the "Course groups" parameter to set up a list of specific courses.

en_us/install_operations/source/configuration/enable_badging.rst
catong, line 508: This checkbox is quite difficult to spot, which is why I added the directions. Not sure what the best option is here. Even for people using screen readers, etc. it would be helpful to know that it is near the top of the page, no? Perhaps I'll just remove "left"? Then it seems weirdly vague.

en_us/install_operations/source/configuration/enable_badging.rst
catong: The whole thing is one configuration, in which you define one or more parameters for awarding cross-course badges. Step 6 says "Next, you create a new course event badge configuration that defines all of
   the course event badges that you want to issue." with a note: "   .. note:: You can create more than one course event badge configuration,
      but only the most recently activated course event badge configuration
      is used."

catong: I think the wording of that note is more clear, so I'll replace the "important" text with the same text as the note in Step 6.

en_us/install_operations/source/configuration/enable_badging.rst
catong: Tried to clarify

en_us/install_operations/source/configuration/enable_badging.rst
catong, line 585: it's an expansion/follow up explanation for step 1

en_us/shared/set_up_course/creating_course_certificates.rst
catong, line 337: Badging is not available for edx.org. Anything about badges is open edx only, so the condition still applies.

catong:
Squashed commits; considering doc complete, but will not merge until feature released on edx.org

catong:
@pbaruah @mattdrayer @bradenmacdonald for a final sanity check of docs against any final implementation changes that might have been made

bradenmacdonald:
same here, pinging @Kelketek

en_us/install_operations/source/configuration/enable_badging.rst
Kelketek, line 186: I made a joke on this during the OpenEdX con last year and no one laughed :(

Kelketek:
Still looks good to me.

catong:
Thanks @Kelketek I think code review and revisions on the code PR is still ongoing. If any changes are made in code that need to be reflected in the code examples in doc, etc., please let me know!

Kelketek:
@catong Will do! Thankfully the notes seem to only have to do with testing, so it's very unlikely any of the workflow or API will change, but I will let you know if anything does manage to on that front.

en_us/install_operations/source/configuration/enable_badging.rst
catong, line 119: Added at lines 133-135

pbaruah:
👍

en_us/open_edx_students/source/SFD_certificates.rst
```diff
@@ -14,6 +14,11 @@ available for courses you take. For information about the certificates that
 your course offers, contact the course team or institution that created your
 course.
 
+In addition, you might be able to earn badges for achievements such as
+completing a course or a series of courses, or for participating in course
+activities. The badges that you can earn depend on how badges have been set up
+for the course platform that you are taking a course on.
+
```
catong, line 21: @bradenmacdonald @Kelketek @pbaruah I just added changes to the Open edX Learner's Guide today. Please review this file - changes introduced to the Learner's Guide are all in this latest commit: added a topic for viewing badges in the user profile and sharing badges from the Accomplishments page.

Draft HTML output: http://draft-open-learners-badgesv2.readthedocs.org/en/latest/SFD_certificates.html#viewing-earned-badges-in-your-profile

As I mentioned to @bradenmacdonald and @pbaruah there is wording in the Mozilla Backpack share dialog left over from web certs badge sharing v1 that refers to "certificates". It would be ideal if we could change the 2 occurrences of "certificates" to "badges" in that dialog.

en_us/open_edx_students/source/SFD_certificates.rst
```diff
@@ -74,36 +79,85 @@ these guidelines.
...
+===============================
+Share Badges From Your Profile
+===============================
+
+To share badges from your profile on the Mozilla Backpack site, follow these
+steps.
+
+#. On your profile, select **Accomplishments**.
+
+#. For the badge that you want to share, select the **Share** icon.
```
bradenmacdonald: Very minor, optional nit: I would consider it a "Share button" and not a share icon.

bradenmacdonald:
👍

en_us/open_edx_students/source/SFD_certificates.rst
bradenmacdonald, line 21: Just reviewed; had one minor comment but it's totally optional, so gave my +1 below.

en_us/open_edx_students/source/SFD_certificates.rst
lamagnifica: +1 "...select **Share**."

pbaruah:
I'll file a ticket to change certificate to badge. 

The learner facing documentation looks good, 👍

catong:
Merging https://github.com/edx/edx-documentation/pull/735

PR #751, opened 2015-12-23
Minute changes found while trying to insert the contents directive

lamagnifica:
DOC-2382 is about making sure each file starts with the contents directive, but expanded to include spell checking, “learners” and “select”. Found a few small changes in the learner’s guides. @catong


lamagnifica:
@srpearce could you look this over? this was not a comprehensive edit, just a few changes I found while looking for the contents directive...

srpearce:
:+1:

PR #761, opened 2016-01-05
Feedback XBlock docs

pmitros:
Adds documentation for the FeedbackXBlock. 

I added text and screenshot, but I did not compile/verify/etc. since historically, docs preferred to rewrite based on something like this regardless. Please let me know if you'd like this brought closer to production-ready with regards to that. 

Should wait for https://github.com/edx/edx-platform/pull/10747 to merge.


catong:
@pmitros Thanks for getting this started. https://openedx.atlassian.net/browse/DOC-2602 tracks the documentation work for FeedbackXBlock.

lamagnifica:
Closing until prioritization is available.

pmitros:
@lamagnifica Hmmm... Looking at this, 10747 is ready for merge. I'm not sure how this fell off the radar.

pmitros:
I'll merge 10747 next week, and then hopefully we can do this one, perhaps? Sorry for letting this stall out.

lamagnifica:
Great, thanks @pmitros. We will definitely need to know the level of support for this XBlock from a product management perspective before we take on doc work, though.

catong:
@marcotuts or @sstack22 Can either of you confirm the plan/support level for the Feedback XBlock?

marcotuts:
@catong I've looked through the history of this and pending some of the feedback noted by Frances I would still consider this not supported. Happy to review again, but I would echo some of the same feedback Frances gave a while back.

pmitros:
@marcotuts Could you please be specific? I thought Frances' feedback was addressed a long time ago. The PR should merge into masters, which has sort of been on hold.

marcotuts:
@pmitros  - It wasn't clear to me that there were any commits after your initial assessment from that PR: 

"A major limitation is that the only way to get data out is through data dumps. This is a little fundamental for the qualitative feedback; future iterations will likely improve this for quantitative. The UX could be better; we based it on how the LTI version of Dropthought looks, but now that Dropthought is no longer around, we could do better. The use of Unicode for faces has upsides and downsides. I think all of these would be good to explore, but not in the MVP." 

In addition to that note, it wasn't clear to me that the feedback Frances and Leslie's provided was addressed in this MVP. Given this context I presumed the XBlock was still an MVP for testing and learning, and thus not provisionally or fully supported.

 If there were any changes to the block since then, would you mind sharing a sandbox / view wehre I could preview the latest version of this XBlock?

pmitros:
@marcotuts 
- UX was redone with Frances, and passed UX/accessibility review. 
- It's possible to view statistics on votes through the instructor view, but qualitative feedback is still data dump. There is no way to do otherwise scalably in the platform as-is.

catong:
Closing this PR.

PR #769, opened 2016-01-08
GitHub has a capital H

singingwolfboy:
All of GitHub's own usage of their name has a capital H. It's even on their [logo usage page](https://github.com/logos). We should respect their preferences about how their name is used.

### DIFF
--- a/en_us/edx_style_guide/source/wordlist.rst
+++ b/en_us/edx_style_guide/source/wordlist.rst
@@ -87,4 +87,6 @@ Word List
        result as the staff view or student view.
    * - XBlock
      - Use instead of "Xblock" or "xBlock".
+   * - GitHub
+     - The "H" is always capitalized.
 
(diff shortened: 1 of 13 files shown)


lamagnifica:
there aren't nearly as many as I expected! Thanks for taking care of this. 
One nice-to-have suggestion: you could add "GitHub" to the wordlist.rst file in the style guide, so that we have a record that this is in fact our preferred way of presenting the name,
:+1:

singingwolfboy:
Added!

en_us/edx_style_guide/source/wordlist.rst
```diff
@@ -87,4 +87,6 @@ Word List
        result as the staff view or student view.
    * - XBlock
      - Use instead of "Xblock" or "xBlock".
+   * - GitHub
+     - The "H" is always capitalized.
```
lamagnifica, line 91: :+1:

PR #772, opened 2016-01-12
Fixes the numbering in the last section

lamagnifica:
@dsjen @pdesjardins @srpearce @catong 

I also fixed a few things that caught my eye, but mostly the fix is to correct the indentation of the bullets in the last section of the page. That, in turn, allows the numbering to render correctly


srpearce:
:+1:

PR #795, opened 2016-01-21
Corrects the heading levels

lamagnifica:
@catong @pdesjardins @srpearce I noticed in the rendered HTML that the wrong heading level was used for the Studio file. This should correct that error.
Tests are running.


pdesjardins:
Looks right to me. :+1:

PR #799, opened 2016-01-22
Remove mention of XBlocks that aren't installed by default

nedbat:


nedbat:
@lamagnifica Sorry, I overlooked that these XBlocks are installed on edx.org, but not by default in Open edX.  We should remove them from the Dogwood release notes.

en_us/open_edx_release_notes/source/dogwood.rst
```diff
@@ -30,30 +30,6 @@ The following new features are included in the Open edX Dogwood release.
 New XBlocks
 ==============
```
lamagnifica, line 25: let's retitle this New LTI XBlock and drop the subheading below

lamagnifica:
The availability and installation of XBlocks is a pretty significant question at this time. 
Should you also be removing the documentation for these two exercise types from the opencoursestaff guide referenced in the links? doc on how to enable them for a course and set them up is pretty useless if they can't install them at the site level...
sigh.

nedbat:
Hmm, I guess we should remove from there also.  I'm really torn about whether edx.org should be documenting these XBlocks, or whether we should be encouraging them to document them well enough that we can just refer to their docs.  Maybe we can get an opinion from @jbarciauskas on Monday.

jbarciauskas:
Based on our conversation this morning :+1: on removing, at least while we sort things out, and potentially link to Confluence page or whatever we decide.

nedbat:
Closing this to make a new PR against the proper branch.

PR #806, opened 2016-01-26
removes duplicated images in incorrect directory

lamagnifica:
@catong I ran into the same problem with a PR I’m trying to merge, discovered that these were duplicated in both this old path and the new /images/ path.


catong:
Ah! I must have missed a clean up step when I moved those images after bringing in the Teams doc. Thanks! :+1:

PR #808, opened 2016-01-26
Adds a redirect within the student and Open student guides for math formatting

lamagnifica:
@pdesjardins @catong @srpearce  do you think this will alleviate distress in people following bookmarks in ongoing courses?
5 minute fix may save a lot of pain
Please sanity check by syncing locally and building, comparing to the old link…


srpearce:
:+1: I was going to suggest this. Thanks!

pdesjardins:
Alternately, we could use the Read the Docs page redirect.

PR #812, opened 2016-01-28
DOC-2583 Configure allowed email patterns for registration

catong:
This PR adds configuration documentation for this feature to the _Installing, Configuring, and Running the Open edX Platform_ guide.
Dev story is OSPR-1023; (already merged) code PR is https://github.com/edx/edx-platform/pull/10943
### Reviewers
- [x] Subject matter expert: @bradenmacdonald
- [ ] Product owner: @griffresch?
- [x] Doc team review: @pdesjardins 
### Testing
- [x] Run ./run_tests.sh without warnings or errors
### Post-review
- [x] Squash commits

### DIFF
--- a/en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
+++ b/en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
@@ -1,0 +1,69 @@
+.. _Configure Allowed Registration Email Patterns:
+
+################################################
+Specifying Allowed Registration Email Patterns
+################################################
+
+This topic describes how to restrict registration on your site by specifying
+which email address patterns are allowed in registration emails.
+
+.. contents::
+   :local:
+   :depth: 1
+
+*********
+Overview
+*********
+
+By default, all email addresses are accepted when learners register for an
+account on your Open edX site. You have the option of restricting registrations
+to learners who use an allowed email address pattern. Doing so can be useful in
+cases where you want to allow only learners who are members of a school,
+organization, or corporation to register and access your courses.
+
+.. note:: Configuring your site using the procedure below only restricts
+   registration to learners whose email addresses match the specified patterns.
+   It does not hide courses from any learners, or prevent access to pages on
+   your site that can be accessed without registration.
+
+
+*********************************
+Configure Allowed Email Patterns
+*********************************
+
+To specify the email patterns that are allowed for registration, follow these steps.
+
+#. Locate the ``lms.env.json`` and ``cms.env.json`` files, which are located
+   one level above the ``edx-platform`` directory. You make the same changes
+   to both files.
+
+#. In the ``lms.env.json`` and ``cms.env.json`` files add the
+   ``REGISTRATION_EMAIL_PATTERNS_ALLOWED`` setting.
+
+   .. code-block:: bash
+
+    "REGISTRATION_EMAIL_PATTERNS_ALLOWED": null
+
+
+   If the value for this setting is ``null``, there are no restrictions, and all
+   email addresses are accepted for registration.
+
+#. Use one or more Python regular expressions to specify the email domains that
+   allowed email addresses must match.
+
+   The following example allows email addresses using the pattern
+   ``example.com`` or ``any.example.com`` to register. It also allows
+   ``school.tld`` addresses, but only if those addresses have a  ``.`` before
+   the ``@`` symbol.
+
+   .. code-block:: bash
+
+     "REGISTRATION_EMAIL_PATTERNS_ALLOWED" = [
+
+        "^.*@(.*\\.)?example\\.com$",
+        "(^\\w+\\.\\w+)@school\\.tld$"
+     ]
+
+#. Save the ``lms.env.json`` and ``cms.env.json`` files.
+
+#. Restart your ``edxapp`` instances.
--- a/en_us/install_operations/source/configuration/index.rst
+++ b/en_us/install_operations/source/configuration/index.rst
@@ -12,6 +12,7 @@ configuration options.
 
    updating_platform
    customize_registration_page
+   config_allowed_regis_emails
    enable_prerequisites
    enable_entrance_exams
    enable_licensing


en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,62 @@
...
+
+
+*********************************
+Configure Allowed Email Patterns
+*********************************
+
+To specify the email patterns that are allowed for registration, you modify a
+setting in the ``lms.env.common.py`` file.
+
+#. In the **Registration** section of the ``lms.env.common.py`` file, locate the
```
bradenmacdonald: I think you are referring to `lms/envs/common.py` here? That file should not be edited. Instead, edits should be made by adding a `REGISTRATION_EMAIL_PATTERNS_ALLOWED` setting to `lms.env.json` (and making the same addition/edit to `cms.env.json` since people could otherwise skip this restriction by registering using Studio.

Since that file is JSON instead of python, it would be formatted like this (note all backslashes need to be doubled):

```
"REGISTRATION_EMAIL_PATTERNS_ALLOWED": [
    ".*@edx\\.org",
    "^.*@opencraft\\.com$"
],
```

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,62 @@
...
+
+     REGISTRATION_EMAIL_PATTERNS_ALLOWED=[
+
+        r'^.*@(.*\.)?example\.com$',
+        r'^(^\w+\.\w+)@school.tld$',
+     ]
+
+#. Save the ``lms.env.common.py`` file.
+
+#. Run database migrations.
```
bradenmacdonald: Not necessary to run any migrations. But on a production instance, you do need to restart the edxapp instances.

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,62 @@
...
+*********************************
+Configure Allowed Email Patterns
+*********************************
+
+To specify the email patterns that are allowed for registration, you modify a
+setting in the ``lms.env.common.py`` file.
+
+#. In the **Registration** section of the ``lms.env.common.py`` file, locate the
+   ``REGISTRATION_EMAIL_PATTERNS_ALLOWED`` setting. By default, the value for
+   this setting is ``None``.
```
bradenmacdonald: Yep, the default is `None`, and it's worth mentioning that `None` means no restrictions (any email will be accepted).

catong:
@bradenmacdonald I've made changes per your review comments. Would you mind taking a final look and especially checking the syntax and regexes that I use in the examples? Thanks!

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,70 @@
...
+Configure Allowed Email Patterns
+*********************************
+
+To specify the email patterns that are allowed for registration, you modify the
+``lms.env.json`` and ``cms.env.json`` files, which are located one level above
+the ``edx-platform`` directory.
+
+#. In the ``lms.env.json`` and ``cms.env.json`` files add the
+   ``REGISTRATION_EMAIL_PATTERNS_ALLOWED`` setting in the ``REGISTRATION``
+   directory.
```
bradenmacdonald: There is no "REGISTRATION directory" in the file.

Here's what the file may look like before making this change:

``` json
...
    "PDF_RECEIPT_TERMS_AND_CONDITIONS": "Here are some fake terms and conditions that say when payment is due and stuff$
    "PERFORMANCE_GRAPHITE_URL": "SetPerformanceGraphiteHostName",
    "PRESS_EMAIL": "press@example.com",
    "PROCTORING_SETTINGS": {},
    "PROFILE_IMAGE_MAX_BYTES": 1048576,
    "PROFILE_IMAGE_MIN_BYTES": 100,
    "REGISTRATION_EXTRA_FIELDS": {
        "city": "hidden",
        "country": "hidden",
        "gender": "optional",
        "goals": "optional",
        "honor_code": "required",
        "level_of_education": "optional",
        "mailing_address": "optional",
        "year_of_birth": "optional"
    },
...
```

and after:

``` json
...
    "PDF_RECEIPT_TERMS_AND_CONDITIONS": "Here are some fake terms and conditions that say when payment is due and stuff$
    "PERFORMANCE_GRAPHITE_URL": "SetPerformanceGraphiteHostName",
    "PRESS_EMAIL": "press@example.com",
    "PROCTORING_SETTINGS": {},
    "PROFILE_IMAGE_MAX_BYTES": 1048576,
    "PROFILE_IMAGE_MIN_BYTES": 100,
    "REGISTRATION_EMAIL_PATTERNS_ALLOWED": [
        ".*@edx\\.org",
        "^.*@opencraft\\.com$"
    ],
    "REGISTRATION_EXTRA_FIELDS": {
        "city": "hidden",
        "country": "hidden",
        "gender": "optional",
        "goals": "optional",
        "honor_code": "required",
        "level_of_education": "optional",
        "mailing_address": "optional",
        "year_of_birth": "optional"
    },
...
```

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,70 @@
...
+
+   .. code-block:: bash
+
+    ############## Registration ##############
+
+    REGISTRATION_EMAIL_PATTERNS_ALLOWED = None
```
bradenmacdonald: Should be `"REGISTRATION_EMAIL_PATTERNS_ALLOWED": null` since this is a JSON file. This will get translated to the Python equivalent, which is what you have above.

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,70 @@
...
+
+
+  If the value for this setting is ``None``, there are no restrictions, and all
```
bradenmacdonald: is `null` or `None`

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,70 @@
...
+#. Replace ``None`` with one or more regular expressions to specify the email
+   domains that allowed email registrations must match.
+
+   The following example allows email addresses using the pattern
+   ``example.com`` or ``any.example.com`` to register, as well as only
+   ``school.tld`` addresses that have a  ``.`` before the ``@`` symbol.
+
+   .. code-block:: bash
+
+     REGISTRATION_EMAIL_PATTERNS_ALLOWED = [
```
bradenmacdonald:
``` json
"REGISTRATION_EMAIL_PATTERNS_ALLOWED": [
    "^.*@(.*\\.)?example\\.com$",
    "(^\\w+\\.\\w+)@school\\.tld$"
]
```

bradenmacdonald:
@catong Closer :-) Left a few tweaks needed.

catong:
@bradenmacdonald 3rd time lucky? Thanks for catching the JSON vs Python issue!

bradenmacdonald:
:+1: 

Thanks @catong !

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,67 @@
+.. _Configure Allowed Registration Email Patterns:
+
+################################################
+Specifying Allowed Registration Email Patterns
```
pdesjardins, line 4: Specifying Registration Email Pattern Restrictions?

Something about the word "allowed" didn't read smoothly to me. Just a thought.

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,67 @@
...
+################################################
+
+This topic describes how to restrict registration on your site by specifying
+which email patterns are allowed in registration emails.
```
pdesjardins: which email address patterns are allowed?

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,67 @@
...
+   :local:
+   :depth: 1
+
+*********
+Overview
+*********
+
+By default, all email addresses are accepted when learners register for an
+account on your Open edX site. You have the option of restricting
+registrations only to learners who use an allowed email address pattern. Doing
```
pdesjardins: Omit "only"?

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,67 @@
...
+so can be useful in cases where you want to only allow learners who are
```
pdesjardins: want to allow only learners who are?

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,67 @@
...
+
+*********************************
+Configure Allowed Email Patterns
+*********************************
+
+To specify the email patterns that are allowed for registration, you modify the
+``lms.env.json`` and ``cms.env.json`` files, which are located one level above
+the ``edx-platform`` directory.
+
+#. In the ``lms.env.json`` and ``cms.env.json`` files add the
```
pdesjardins, line 40: I'm only mentioning this because I have been trying to adhere to the style guide strictly. It requests a specific introduction paragraph:

> To introduce an ordered list of procedural steps, use “To do this, follow these steps.” Note the complete sentence and terminal period.

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,67 @@
...
+
+   .. code-block:: bash
+
+    "REGISTRATION_EMAIL_PATTERNS_ALLOWED": null
+
+
+   If the value for this setting is ``null``, there are no restrictions, and all
+   email addresses are accepted for registration.
+
+#. Use one or more regular expressions to specify the email domains that allowed
```
pdesjardins: It might be important to specify which regular expression syntax is accepted here. I'm guessing that it's interpreted by the Python re module (https://docs.python.org/2/library/re.html). Some people might be more used to different UNIX tool syntaxes and try to use them.

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,67 @@
...
+   email registrations must match.
+
+   The following example allows email addresses using the pattern
+   ``example.com`` or ``any.example.com`` to register, as well as only
```
pdesjardins: Breaking into two sentences might make this clearer.

The following example allows email addresses using the pattern `example.com` or `any.example.com` to register. It also allows `school.tld` addresses but only if those addresses have a  `.` before the `@` symbol.

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,67 @@
...
+   ``school.tld`` addresses that have a  ``.`` before the ``@`` symbol.
+
+   .. code-block:: bash
```
pdesjardins, line 59: Just out of curiosity, why use BASH and not JSON? It's a JSON example, right? Maybe Sphinx doesn't accept JSON as a code-block type?

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
```diff
@@ -0,0 +1,67 @@
...
+
+     "REGISTRATION_EMAIL_PATTERNS_ALLOWED" = [
+
+        "^.*@(.*\\.)?example\\.com$",
+        "(^\\w+\\.\\w+)@school\\.tld$"
+     ]
+
+#. Save the ``lms.env.json`` and ``cms.env.json`` files.
```
pdesjardins, line 67: Maybe this goes without saying, but I assume the setting for REGISTRATION_EMAIL_PATTERNS_ALLOWED must be the same in both .json configuration files. No? If they were different, I wonder what would happen. That might come up if someone updated one file and forgot about the other.

pdesjardins:
I added some comments, looks good! :+1:

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
catong: Yeah, I wondered about that. @bradenmacdonald should we specify which flavor regex is accepted?

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
catong, line 59: @pdesjardins I did start out with JSON code formatting, but decided to go back to BASH so that the code example (colors, etc.) look consistent with other examples in this guide. It could be part of the JIRA story to correct code examples to also correct the example formatting.

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
catong, line 67: I've rewritten the first couple of steps per your review comments and I think this comment should be addressed by my changes.

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
catong, line 4: I'd like to keep as is for a couple of reasons - it matches the configuration setting name, and I find "Registration + Restrictions" more difficult to parse than "Allowed".

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
bradenmacdonald: Yes, good point. It is the Python syntax.

catong:
@pdesjardins Addressed your review comments. @bradenmacdonald Can you please chime in re supported regex syntax? Thanks!

en_us/install_operations/source/configuration/config_allowed_regis_emails.rst
bradenmacdonald, line 67: They can be different, and it would mean that different rules would be enforced on the LMS vs the CMS. This is an unfortunate downside of how edx-platform splits all of its configuration into two files without a shared config section.

catong, line 67: @bradenmacdonald As you point out, it's the case with all our configs that you could configure the LMS differently from the CMS, but generally you would not want to. @pdesjardins We don't get into "configure differently" for any other existing features; I'm not going to mention that option here either. We only present the procedure for the most common use case.

catong:
@bradenmacdonald and @pdesjardins Made final tweaks. I already have thumbs from both of you but that was before some of the final comments. Good to go?

pdesjardins:
Yes, looks good to me. :+1:

bradenmacdonald:
Yep, :+1: stands :) Thanks @catong

PR #813, opened 2016-01-29
Remove unlicensed image 

lamagnifica:
@xcompass this PR removes the original screen shot of the image and leaves only the obscured version.
@pdesjardins or @catong sanity check please.


pdesjardins:
Looks good to me. I don't see any references to that image anywhere in our doc source. Wasn't needed anyway? :+1:

xcompass:
:+1: @pdesjardins It's replaced by obscured version.

lamagnifica:
Yes, @pdesjardins, in my original PR I chose to add both this (original) version, and also a version that obscured the unlicensed photo. Now that we know that we will not be using the photo, we can remove the screen shot that shows it clearly.

PR #821, opened 2016-02-02
Pin a dependency to keep develop.sh working

nedbat:


nedbat:
@pdesjardins this fixes things.

pdesjardins:
:+1:  Maybe drop in a comment to explain the change? So we can remove the pin later?

pdesjardins:
Oh, please ignore the comment, comment. :+1:

PR #824, opened 2016-02-02
Added a shell script that builds Read the Docs projects

pdesjardins:
## [DOC-2651](https://openedx.atlassian.net/browse/DOC-2651)

I am adding a shell script that will invoke the REST APIs that build the Read the Docs projects for our publicly visible documents. I chose to add it in a new top-level directory because it doesn't belong with the content in en_us and the shared directory is a specialized Python module. I would be happy to put this somewhere else, if that is preferred.
### Reviewers
- [ ] Doc team review (sanity check): @lamagnifica @catong @srpearce 
### Testing
- [x] I have run this script myself several times. It prompts builds on RTD in the same way that clicking the build button from a project dashboard does.
### Post-review
- [ ] Squash commits

### DIFF
--- a/utilities/build-rtd-documents.sh
+++ b/utilities/build-rtd-documents.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+# This script invokes the REST API that builds edX Read the Docs projects
+# for publicly visible documents.
+# The script requires the curl HTTP client (installed by default in most
+# Linux and Mac OSes) and a connection to the internet. Verify that you
+# have curl installed by invoking the "curl --version" command in a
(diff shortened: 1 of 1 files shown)


catong:
@pdesjardins It makes sense to me that this goes at the level above en_us.

pdesjardins:
@catong Do you mean that you think the script should be moved up a directory level? Or that it makes sense to have it in the utilities directory (which is a sibling to en_us)?

catong:
@pdesjardins Oops, I meant that the script is fine where you put it, in the utilities directory.

PR #825, opened 2016-02-02
Ned/dogwood docs

nedbat:
The latest Dogwood docs, for the dogwood branch.


nedbat:
@lamagnifica does this look right?

lamagnifica:
yes @nedbat

PR #826, opened 2016-02-02
Remove mention of XBlocks that aren't installed by default

nedbat:


nedbat:
@lamagnifica I found the XBlock removal.  We had commented and tweaked it together, and then forgotten about it!

lamagnifica:
:+1:

PR #829, opened 2016-02-02
Fix visible release dates

catong:
Fix visible release dates in the release notes


catong:
merging https://github.com/edx/edx-documentation/pull/829

PR #925, opened 2016-03-16
Adds more practical guidance on when it's appropriate to use offscreen text

cptvitamin:
@catong can you review this for grammatical correctness and readability? 
@clrux do you think this guidance is helpful and complete?

### DIFF
--- a/en_us/developers/source/conventions/accessibility.rst
+++ b/en_us/developers/source/conventions/accessibility.rst
@@ -494,10 +494,12 @@ the Accessibility API.
 
 To visibly hide content that should be read by screen readers, edX makes a CSS
 ``class="sr"`` available to expose content only to non-visual users. This is
-achieved by displaying the content beyond the bounds of the viewport so that it 
-is not displayed visually, but still accessible in the DOM.  This is sometimes
-referred to as displaying content "offscreen". In the following example, a
-visual user sees only the X, while a screen reader user hears only "Close".
+achieved by displaying the content beyond the bounds of the viewport, or
+clipping the content to a single pixel. These techniques remove any visible
+trace of the element from the page, while still leaving it accessible to screen
+reader users. This is often referred to as displaying content "offscreen". In
+the following example, a visual user sees only the X, while a screen reader
+user hears only "Close".
 
 ::
 
@@ -507,45 +509,48 @@ visual user sees only the X, while a screen reader user hears only "Close".
   </a>
   
 The choice to show or hide content for a specific audience should not be taken
-lightly. Extensive use of offscreen content often indicates poor design or 
-accessibility. The following questions should help you decide if one of the above 
-techniques is appropriate or not:
+lightly. Extensive use of offscreen content can reduce accessibility, and is
+often an indicator of a user experience that relies too heavily on visual
+context. The following questions should help you decide whether it is
+appropriate to hide content from screen readers or display it offscreen.
 
 Would all users benefit from the content displayed offscreen?
-  If the additional content considered for display offscreen could also be useful for 
-  visual users as well, consider finding a way to make the content work visually.
+  If the content you are considering displaying offscreen might be useful not only
+  for non-visual users but other users too, find a way to make the content work
+  visually, and expose it for all users.
   
-Is important context being provided only visually?
-  Consider a standard sidebar navigation. It is common to visually indicate the
-  current page or section by differentiating it visually from other pages or
-  sections in the navigation.  To a visual user, it is clear that the item in 
-  the list that looks different than all the others is the page they are currently
-  viewing. Offscreen text is an effective way to make this visual context available 
-  to non-visual users. The following example demonstrates this.  *Note the non-breaking
-  space in the offscreen text span.  Without it, a screen reader would read this text
-  as "About UsCurrent page."*
+Are you using only visual cues to provide important context?
+  In standard sidebar navigation, it is common practice to indicate the user's
+  current page or section by differentiating it visually from other pages or sections
+  in the sidebar. To visual users, it is clear that the item in the list that looks
+  different than all the others is the page that they are currently viewing. You can
+  make this visual context available to non-visual users with offscreen text, as
+  demonstrated in the following example.
   
 ::
 
   <a href="/" class="inactive">Home</a>
   <a href="about/" class="active">About Us<span class="sr">&nbsp;Current page</span></a>
   
-Does the content displayed offscreen contain any interactive elements?
-  Offscreen content should never contain interactive elements, like links, 
-  buttons, or form inputs.  This would have a negative impact on sighted 
-  keyboard-only users who require visual focus indicators to understand what 
-  element has focus and will be the target of keyboard events.
+.. note:: In the code example above, the non-breaking space prevents a screen reader 
+  from reading the text as “About UsCurrent Page”.
   
-Is offscreen content being used to label an interactive element?
-  It may be tempting to use offscreen text to improve the usability of an interactive
+Does the content displayed offscreen contain any interactive elements?
+  Never include interactive elements such as links, buttons, or form inputs, in
+  offscreen content. Doing so negatively impacts sighted keyboard-only users, who
+  require visual focus indicators to understand what element has focus and will be
+  the target of keyboard events.
+  
+Are you including interactive elements in offscreen content?
+  It can be tempting to use offscreen text to improve the usability of an interactive
   element for non-visual users. Offscreen text is included in the Accessibility API
-  which is used by screen readers.  However, screen readers are not the only 
-  assistive technology that use the Accessibility API. Speech input software will
-  also use the Accessibility API to identify interactive controls.  In the 
-  following example, a screen reader user will hear "Type your First Name", but a
-  sighted user will see only "First Name."  A user who relies on speech input to
+  which is used by screen readers. However, screen readers are not the only 
+  assistive technology that use the Accessibility API. Speech input software also
+  uses the Accessibility API to identify interactive controls. In the 
+  following example, screen reader users will hear "Type your First Name", but
+  sighted users will see only "First Name." Users who rely on speech input to
   interact with their computer will move focus to this element by saying "Focus on
-  First Name" (the visual label).  However, the Accessible label for this element is
+  First Name" (the visual label). However, the accessible label for this element is
   "Type your First Name."
   
 ::
--- a/en_us/developers/source/conventions/accessibility.rst
+++ b/en_us/developers/source/conventions/accessibility.rst
@@ -514,18 +514,20 @@ often an indicator of a user experience that relies too heavily on visual
 context. The following questions should help you decide whether it is
 appropriate to hide content from screen readers or display it offscreen.
 
-Would all users benefit from the content displayed offscreen?
-  If the content you are considering displaying offscreen might be useful not only
-  for non-visual users but other users too, find a way to make the content work
-  visually, and expose it for all users.
+* Would all users benefit from the content displayed offscreen?
+
+  * If the content you are considering displaying offscreen might be useful not only
+    for non-visual users but other users too, find a way to make the content work
+    visually, and expose it for all users.
   
-Are you using only visual cues to provide important context?
-  In standard sidebar navigation, it is common practice to indicate the user's
-  current page or section by differentiating it visually from other pages or sections
-  in the sidebar. To visual users, it is clear that the item in the list that looks
-  different than all the others is the page that they are currently viewing. You can
-  make this visual context available to non-visual users with offscreen text, as
-  demonstrated in the following example.
+* Are you using only visual cues to provide important context?
+  
+  * In standard sidebar navigation, it is common practice to indicate the user's
+    current page or section by differentiating it visually from other pages or sections
+    in the sidebar. To visual users, it is clear that the item in the list that looks
+    different than all the others is the page that they are currently viewing. You can
+    make this visual context available to non-visual users with offscreen text, as
+    demonstrated in the following example.
   
 ::
 
@@ -535,23 +537,25 @@ Are you using only visual cues to provide important context?
 .. note:: In the code example above, the non-breaking space prevents a screen reader 
   from reading the text as “About UsCurrent Page”.
   
-Does the content displayed offscreen contain any interactive elements?
-  Never include interactive elements such as links, buttons, or form inputs, in
-  offscreen content. Doing so negatively impacts sighted keyboard-only users, who
-  require visual focus indicators to understand what element has focus and will be
-  the target of keyboard events.
+* Does the content displayed offscreen contain any interactive elements?
+
+  * Never include interactive elements such as links, buttons, or form inputs, in
+    offscreen content. Doing so negatively impacts sighted keyboard-only users, who
+    require visual focus indicators to understand what element has focus and will be
+    the target of keyboard events.
   
-Are you including interactive elements in offscreen content?
-  It can be tempting to use offscreen text to improve the usability of an interactive
-  element for non-visual users. Offscreen text is included in the Accessibility API
-  which is used by screen readers. However, screen readers are not the only 
-  assistive technology that use the Accessibility API. Speech input software also
-  uses the Accessibility API to identify interactive controls. In the 
-  following example, screen reader users will hear "Type your First Name", but
-  sighted users will see only "First Name." Users who rely on speech input to
-  interact with their computer will move focus to this element by saying "Focus on
-  First Name" (the visual label). However, the accessible label for this element is
-  "Type your First Name."
+* Are you including interactive elements in offscreen content?
+
+  * It can be tempting to use offscreen text to improve the usability of an interactive
+    element for non-visual users. Offscreen text is included in the Accessibility API
+    which is used by screen readers. However, screen readers are not the only 
+    assistive technology that use the Accessibility API. Speech input software also
+    uses the Accessibility API to identify interactive controls. In the 
+    following example, screen reader users will hear "Type your First Name", but
+    sighted users will see only "First Name." Users who rely on speech input to
+    interact with their computer will move focus to this element by saying "Focus on
+    First Name" (the visual label). However, the accessible label for this element is
+    "Type your First Name."
   
 ::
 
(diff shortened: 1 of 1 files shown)


en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
 the Accessibility API.
 
 To visibly hide content that should be read by screen readers, edX makes a CSS
-``class="sr"`` available to expose content only to screen reader users. In the
-following example, a sighted user sees only the X, while a screen reader user
-hears only "Close".
+``class="sr"`` available to expose content only to non-visual users. This is
+achieved by displaying the content beyond the bounds of the viewport so that it
```
downzer0: The updated method for using `.sr` classes doesn't actually position it outside the bounds of the viewport. Instead, the "wrapper" or "container" with the class is given the dimensions of a single pixel, clipped, and any overflow is hidden. Not sure how to say that though, without being too technical.

downzer0: Furthermore, if we sync up the descriptions to match the technicality, any use of "offscreen" would need to change.

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+Is important context being provided only visually?
+  Consider a standard sidebar navigation. It is common to visually indicate the
+  current page or section by differentiating it visually from other pages or
+  sections in the navigation.  To a visual user, it is clear that the item in 
+  the list that looks different than all the others is the page they are currently
+  viewing. Offscreen text is an effective way to make this visual context available 
+  to non-visual users. The following example demonstrates this.  *Note the non-breaking
+  space in the offscreen text span.  Without it, a screen reader would read this text
+  as "About UsCurrent page."*
+
```
downzer0, line 531: I'm wondering if this can be simplified to "Is the content meaningful and does it have purpose?"

downzer0:
@cptvitamin Definitely helpful. Does this coincide with any actual examples? If not, should we include a short Fiddle in this documentation? Also, I had a few questions about terminology vs technical mechanics.

en_us/developers/source/conventions/accessibility.rst
cptvitamin: @clrux It's a concept/paradigm that is well known in the industry (all articles written on the topic refer to this technique as offscreen).  I think referring to the implementation details is too pedantic and will not make it any easier to understand.

cptvitamin:
@clrux there are quite a few code examples in there already.  Do you think we need more?

en_us/developers/source/conventions/accessibility.rst
cptvitamin, line 531: It's unclear to me which line you are referring to.

downzer0:
Sorry, I meant functional, working examples that are linked to from this document. Is that too complicated for something like this?

en_us/developers/source/conventions/accessibility.rst
downzer0, line 531: This applies to 514 - 526.

cptvitamin, line 531: If you were to venn diagram that statement, you could easily have content that is "meaningful and has purpose" inside the circle for offscreen content (non-visual users).  How about "Is the offscreen content generally useful for all users, not just non-visual users?"

downzer0, line 531: I think that is more clear, yes.

cptvitamin:
@clrux we haven't done that for any of the other guidelines in this document.  While it would be helpful, I feel like we have provided enough info, context, and code examples for developers to use it effectively.

cptvitamin:
@catong would you have a minute to look this over and provide feedback?

catong:
@cptvitamin Sorry for the delay; I can take a look starting this afternoon. Thanks!

en_us/developers/source/conventions/accessibility.rst
catong: @clrux @cptvitamin I don't have any context other than your discussion here, but would it be more accurate to refer to both methods? Something like: 
"This is achieved either by positioning the content outside the bounds of the viewport, or by setting the element to a single pixel in height and performing clipping using CSS. These methods remove any visible trace of the element from the page, while still leaving it available for screen readers."

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
 
   <a href="#">
-  <span aria-hidden="true">X</span>
-  <span class="sr">Close</span>
+    <span aria-hidden="true">X</span>   <!-- hidden from screen reader users -->
+    <span class="sr">Close</span>       <!-- exposed only to screen reader users -->
   </a>
+  
+The choice to show or hide content for a specific audience should not be taken
+lightly. Extensive use of offscreen content often indicates poor design or
```
catong: Where we say "offscreen" should we say "offscreen or hidden"?

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
-  <span aria-hidden="true">X</span>
-  <span class="sr">Close</span>
+    <span aria-hidden="true">X</span>   <!-- hidden from screen reader users -->
+    <span class="sr">Close</span>       <!-- exposed only to screen reader users -->
   </a>
+  
+The choice to show or hide content for a specific audience should not be taken
+lightly. Extensive use of offscreen content often indicates poor design or 
+accessibility. The following questions should help you decide if one of the above 
+techniques is appropriate or not:
```
catong: Not clear what "one of the above techniques" refers to - the subheading levels?

en_us/developers/source/conventions/accessibility.rst
cptvitamin: @catong that may cause confusion since offscreen and hidden are different things.  Hidden generally means hidden from all users and is achieved by the CSS properties `display: none`  or `visibility: hidden` .  Offscreen implies that the content is still being displayed, but beyond the bounds of the viewport, where it can be accessed by AT, but not visual users.

en_us/developers/source/conventions/accessibility.rst
cptvitamin: @catong I was referring to the code sample above which demonstrates 2 techniques:  `aria-hidden=true` to hide from screen readers, or `class=sr` to hide visually but expose to screen readers.  Any suggestions for how to make that more clear?  I can say "The following questions should help you decide if hiding content from screen readers, or displaying it offscreen, is appropriate"

catong: @cptvitamin I like your suggested rewrite. Please end the sentence in a period rather than a colon.

en_us/developers/source/conventions/accessibility.rst
catong: OK

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+
+Would all users benefit from the content displayed offscreen?
```
catong: These questions would work as bulleted items. Asterisk followed by a space gets you the bullet.

catong: For the paragraph(s) following each bullet question, indent to the same level as the first letter after the asterisk.

en_us/developers/source/conventions/accessibility.rst
catong: @cptvitamin A thought about the wording on line 510. Perhaps soften perceived criticism by changing "often indicates poor design" to "can indicate poor design"? Also, can you clarify what you mean by "Extensive use of offscreen content [can indicate] poor... accessibility"?

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+  If the additional content considered for display offscreen could also be useful for
```
catong: Strike "additional". Per Chris's comment below, perhaps "If the content you are considering displaying offscreen might be useful not only for non-visual users but other users too, find a way to make the content work visually, and expose it for all users."

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+  If the additional content considered for display offscreen could also be useful for 
+  visual users as well, consider finding a way to make the content work visually.
+  
+Is important context being provided only visually?
+  Consider a standard sidebar navigation. It is common to visually indicate the
+  current page or section by differentiating it visually from other pages or
+  sections in the navigation.  To a visual user, it is clear that the item in 
+  the list that looks different than all the others is the page they are currently
+  viewing. Offscreen text is an effective way to make this visual context available 
+  to non-visual users. The following example demonstrates this.  *Note the non-breaking
```
catong: Format as a note by entering `.. note::` followed by a space then your note text. See suggestion in comment above.

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+lightly. Extensive use of offscreen content often indicates poor design or 
+accessibility. The following questions should help you decide if one of the above 
+techniques is appropriate or not:
+
+Would all users benefit from the content displayed offscreen?
+  If the additional content considered for display offscreen could also be useful for 
+  visual users as well, consider finding a way to make the content work visually.
+  
+Is important context being provided only visually?
+  Consider a standard sidebar navigation. It is common to visually indicate the
```
catong: Some suggested wording: 
"In standard sidebar navigation, it is common practice to indicate the user’s current page or section by differentiating it visually from other pages or sections in the sidebar. To visual users, it is clear that the item in the list that looks different than all the others is the page that they are currently viewing. You can make this visual context available to non-visual users with offscreen text, as demonstrated in the following example. 

[code example]

.. note:: In the code example above, the non-breaking space prevents a screen reader from reading the text as “About UsCurrent Page”.

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+  space in the offscreen text span.  Without it, a screen reader would read this text
+  as "About UsCurrent page."*
+  
+::
+
+  <a href="/" class="inactive">Home</a>
+  <a href="about/" class="active">About Us<span class="sr">&nbsp;Current page</span></a>
+  
+Does the content displayed offscreen contain any interactive elements?
+  Offscreen content should never contain interactive elements, like links,
```
catong: "like" -> "such as"

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+  as "About UsCurrent page."*
+  
+::
+
+  <a href="/" class="inactive">Home</a>
+  <a href="about/" class="active">About Us<span class="sr">&nbsp;Current page</span></a>
+  
+Does the content displayed offscreen contain any interactive elements?
+  Offscreen content should never contain interactive elements, like links, 
+  buttons, or form inputs.  This would have a negative impact on sighted
```
catong: Only a single space between sentences.

en_us/developers/source/conventions/accessibility.rst
catong: Suggested rewrite: "Never include interactive elements such as links, buttons, or form inputs, in offscreen content. Doing so negatively impacts signed keyboard-only users, who require..."

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+The choice to show or hide content for a specific audience should not be taken
+lightly. Extensive use of offscreen content often indicates poor design or 
+accessibility. The following questions should help you decide if one of the above 
+techniques is appropriate or not:
+
+Would all users benefit from the content displayed offscreen?
+  If the additional content considered for display offscreen could also be useful for 
+  visual users as well, consider finding a way to make the content work visually.
+  
+Is important context being provided only visually?
```
catong: Suggestion: "Are you using only visual cues to provide important context?"

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+  to non-visual users. The following example demonstrates this.  *Note the non-breaking
+  space in the offscreen text span.  Without it, a screen reader would read this text
+  as "About UsCurrent page."*
+  
+::
+
+  <a href="/" class="inactive">Home</a>
+  <a href="about/" class="active">About Us<span class="sr">&nbsp;Current page</span></a>
+  
+Does the content displayed offscreen contain any interactive elements?
```
catong: Suggestion: "Are you including interactive elements in offscreen content?"

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+  Offscreen content should never contain interactive elements, like links, 
+  buttons, or form inputs.  This would have a negative impact on sighted 
+  keyboard-only users who require visual focus indicators to understand what 
+  element has focus and will be the target of keyboard events.
+  
+Is offscreen content being used to label an interactive element?
```
catong: Suggestion: "Are you using offscreen content to label interactive elements?"

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+  It may be tempting to use offscreen text to improve the usability of an interactive
```
catong: Change "may" (implies permission) to either "can" or "might"

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+  element for non-visual users. Offscreen text is included in the Accessibility API
+  which is used by screen readers.  However, screen readers are not the only 
+  assistive technology that use the Accessibility API. Speech input software will
+  also use the Accessibility API to identify interactive controls.  In the
```
catong: "will also use" -> "also uses"

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+  keyboard-only users who require visual focus indicators to understand what 
+  element has focus and will be the target of keyboard events.
+  
+Is offscreen content being used to label an interactive element?
+  It may be tempting to use offscreen text to improve the usability of an interactive
+  element for non-visual users. Offscreen text is included in the Accessibility API
+  which is used by screen readers.  However, screen readers are not the only 
+  assistive technology that use the Accessibility API. Speech input software will
+  also use the Accessibility API to identify interactive controls.  In the 
+  following example, a screen reader user will hear "Type your First Name", but a
```
catong: Change to plural: "screen reader users will hear... but sighted users will see only..."
"Users who rely on speech input" 
"with their computer" -> "with their computers"

en_us/developers/source/conventions/accessibility.rst
```diff
@@ -493,17 +493,66 @@ encumber a different audience. For instance, a **Close** button that looks like
...
+  sighted user will see only "First Name."  A user who relies on speech input to
+  interact with their computer will move focus to this element by saying "Focus on
+  First Name" (the visual label).  However, the Accessible label for this element is
```
catong: lowercase "A" in "accessible"?

catong:
@cptvitamin I've finished my editing pass. I'd be happy to take a final look after edits and formatting fixes, etc. - let me know if you'd help with formatting.

cptvitamin:
@catong I addressed all of your comments except for the conversion of definition lists to unordered or bulleted lists.  I feel like the content is too long to be considered a list item. Also, there would be two distinct pieces of content (question and answer) contained in a single list item.  I think the definition list is the proper markup here.

catong:
@cptvitamin I see your point about not converting your questions to list items. However, because we don't use the definition format anywhere else in the document (I think) the hanging indents are inconsistent from all other content. I also find it difficult to read due to its reduced line spacing. I think the bullets will work fine even though the items are not technically list items. What do you think, can I change your mind? :-) I can provide a draft document but here are some screenshots.

We've used bullets previously for Q&A, and it looks like this.

qaexample

The definition paragraphs look like this.

definitionformat

catong:
@cptvitamin Draft HTML output is here: http://draft-devguide-a11y-offscreen.readthedocs.org/en/latest/conventions/accessibility.html#hide-or-expose-content-to-targeted-audiences

cptvitamin:
@catong I have reluctantly conceded.  Before I retreat, I feel obligated to publicly state that we should never base our decisions for markup on visual appearance.  The list item examples cited above are also not ideal candidates for list and nested-list markup.  In the future we should adjust the definition list line spacing defined in our ReadTheDocs instance, and convert these to definition lists.  For now, I think having the information in this doc is more important.

catong:
@cptvitamin I also realized that we should not be applying formatting based on appearance rather than function. I'm not against the definition formatting, but can you help me understand why they are definitions? Typically we would use "definitions" in something like a glossary where you have a short term (one to three words) followed by a definition for that term. I can't quite see why these lists of questions and their corresponding answers can't be considered to be unordered lists. I think it works for equivalent situations earlier in this document.

cptvitamin:
@catong i used the wrong nomenclature and apologize.  Typically used for definitions (glossary) the `<dl>` element is often referred to as the definition list element. Technically, it's a "description" list.  The [HTML5](https://www.w3.org/TR/html5/grouping-content.html#the-dl-element) spec defines its usage as follows:

> an association list consisting of zero or more name-value groups (a description list) [...] Name-value groups may be terms and definitions, metadata topics and values, questions and answers, or any other groups of name-value data.

The `<ul>` element, or unordered list is defined as

> a list of items, where the order of the items is not important 

When I think of lists, I think of very concise items or phrases.  The examples cited above seem to me to be more questions and answers or metadata topics and values.

PR #946, opened 2016-03-26
Adds ally requirement for Office Mix

lamagnifica:


openedx-webhooks:
## [DOC-2552](https://openedx.atlassian.net/browse/DOC-2552)

Adds info on adding captions to mixes, which are required for time-based content in edx.org and Edge courses.

### Date Needed

As soon as feasible

### Reviewers

Possible roles follow. PR submitter checks the boxes after each reviewer
finishes and gives :+1:. 
- [ ] Subject matter expert: @cptvitamin 
- [ ] Subject matter expert: 
- [ ] Doc team review (sanity check): @catong @pdesjardins @srpearce 

### Testing

### Post-review
- [ ] Squash commits

en_us/shared/exercises_tools/office_mix.rst
```diff
@@ -99,6 +99,22 @@ video files in your course, see :ref:`Working with Video Components`.
   For more information about video processing for courses that run on the
   edx.org site, see :ref:`partnercoursestaff:Video Processing Overview`.
 
+=============================================
+Adding Captions to Office Mix Presentations
+=============================================
+
+When you design a mix, be sure to consider whether it includes time-based media
```
cptvitamin: @lamagnifica thanks for taking this on!  My only concern here is that by using the technical terminology and language for when captions are required makes it sound (to those not familiar with terms like time-based media) like it is not likely, or less likely that the requirement will be triggered (or at least too easy to dismiss).   I can't think of any use cases for Office Mix that would not require captioning.  Can we say something like "Because Office Mix Presentations often include video and/or audio content, they are likely to qualify as time-based media and require captions or a transcript...."  Something like that?  This approach makes the user more likely to make the connection that captions are required, and educates them that audio and video content is the same as time-based media.

lamagnifica: that makes perfect sense, @cptvitamin thanks!

lamagnifica:
@cptvitamin reworded

cptvitamin:
:+1:

PR #947, opened 2016-03-26
Blurs out names, usernames, email addresses

lamagnifica:


openedx-webhooks:
## [DOC-288](https://openedx.atlassian.net/browse/DOC-288)

Blurs PII.

### Reviewers

Possible roles follow. PR submitter checks the boxes after each reviewer
finishes and gives :+1:. 
- [ ] Subject matter expert: 
- [ ] Subject matter expert: 
- [x] Doc team review (sanity check): @srpearce @catong @pdesjardins 
- [ ] Product review:
- [ ] Partner support: 
- [ ] PM review: 

### Testing

### Post-review
- [x] Squash commits

catong:
@lamagnifica A couple of small suggestions: 
- in /shared/students/Images/Mob_WiFiSetting.png - it's still possible to read Sylvia's name. Could the blurring be increased?
- in shared/students/Images/SFD_VerifyID_NonRoman.png - Is there any point in blurring the roman name if we don't also blur the non-roman name? Maybe use a very slight blurring for the non-roman name so that readers can see that they are non-roman characters, but can't make them out. Also, was it intentional that you left the first letter of the roman name unblurred?

lamagnifica:
@catong I'm going to search for uses of that Mob_WiFiSetting.png file. I didn't think of it before, but it seems likely it was just a trial image and not used...

lamagnifica:
ok, I found 12 "Mob_" images that aren't use, build is clean after deleting them. Bombs away.

lamagnifica:
@catong there's my non-global thinking catching me up... new commit with that SFD_VerifyID_NonRoman.png re-blurred...

PR #952, opened 2016-03-28
Adds an example of rewriting into active voice

lamagnifica:


openedx-webhooks:
@srpearce @catong @pdesjardins to revive my assigned goal for FY2016 of getting the style guide in publishable shape, I'm going to be adding little bits here and there as I think of examples and have a minute or two. Most won't have a DOC story, but I'll also be pulling from the backlog. 
Oh, so anyway, the call to action is, if one or all of you want to take a look at these, please do. I'll tag all three of you as I move forward, but for anything potentially controversial I'll bring it up for discussion in standup or other team meetings first.

PR #965, opened 2016-04-04
Companion to PR 12027

lamagnifica:
@pdesjardins @catong @srpearce  this brings the example in the OLX guide in line with Vedran's fix in
https://github.com/edx/edx-platform/pull/12027/files


srpearce:
:+1:

PR #973, opened 2016-04-08
Corrects an oversight in using file URLs section

lamagnifica:
@srpearce @catong @pdesjardins 
Shelby had a question about sizes for uploaded files, so I was looking at this section and noticed an odd line or two. 

other tweaks made also. See http://edx.readthedocs.org/projects/edx-partner-course-staff/en/latest/course_assets/course_files.html#file-size

### DIFF
--- a/en_us/shared/course_assets/course_files.rst
+++ b/en_us/shared/course_assets/course_files.rst
@@ -31,7 +31,6 @@ use the Files & Uploads page to add these files to your course. If you are an
 edX partner and you need to store large data sets, contact your edX program
 manager.
 
-
 .. _Upload a File:
 
 *******************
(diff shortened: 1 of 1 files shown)


en_us/shared/course_assets/course_files.rst
```diff
@@ -78,30 +77,23 @@ The **Files & Uploads** page refreshes to show the uploaded file.
...
 ********************************************
 
-After you upload a file, you can use the URLs listed for it. On the **Files &
-Uploads** page, the **URL** column lists a Studio URL and web URL for each
-file.
-
-Use the web URL if you want to perform either of the following actions.
-Add a file or image to a bulk email message...
-Provide a link to the file or image from outside the course.
+To use an uploaded file for your course, you add a reference to one of the URLs
```
pdesjardins, line 80: I'm not overly worried about long lines, but do these need the Sublime line length plugin applied to them?

pdesjardins:
Looks good! :+1:

en_us/shared/course_assets/course_files.rst
lamagnifica, line 80: um. I used the line length plugin. I think that this line in 79 characters long (+ trailing space). My line length is set to 80 -- is yours set differently?

pdesjardins, line 80: Sorry, it must have been the view in my browser. I was misinterpreting the
line wrap I saw. Now it makes sense.

Peter

PR #975, opened 2016-04-11
Adds xref to all waffle switches for Insights

lamagnifica:


dsjen:
Thank you!  :+1:

brianhw:
If Dennis says it's good then it's good! :+1:

PR #979, opened 2016-04-14
Fix broken links

srpearce:
@lamagnifica @pdesjardins @catong This PR fixes the broken links that a different PR introduced. It just needs a sanity check.


catong:
Green is pretty :-) :+1:

PR #989, opened 2016-04-21
A little more prereq ideas from Joel

nedbat:


nedbat:
@jbarciauskas Based on your comments.

jbarciauskas:
👍

PR #990, opened 2016-04-22
Help and feedback links instead of GitHub.

nedbat:
A proof-of-concept for adding "Need Help?" and "Feedback" links to the top of each page.  

The Need Help links can point to difference links, depending on the audience of the book.  Here there are three different audiences (Open_edX, Partners, and unspecified), but that can be changed.  The actual links are just placeholders, and we'll need to work with Ed Services and Open edX to decide where they should really go.

The Feedback link goes to a sample Google Form that Peter made, with an indication of the project and page included automatically.

### DIFF
--- a/shared/conf.py
+++ b/shared/conf.py
@@ -41,6 +41,9 @@ if on_rtd:
     html_context["github_project"] = 'edx-documentation'
 
 FEEDBACK_FORM_FMT = "https://docs.google.com/forms/d/1gGBVXUS84Q8jVEJdlpYYIkOQmaCg5JnzBTG1x4ASWKM/viewform?entry.930288665&entry.672882205={url}"
+OPEN_HELP = "http://open.edx.org/help"              # JUST A PLACEHOLDER
+PARTNER_HELP = "http://partners.edx.org/help"       # DON'T FREAK, JUST ANOTHER PLACEHOLDER
+GENERIC_HELP = "http://help.com"                    # OBVIOUSLY A JOKE
 
 def feedback_form_url(project, page):
     """Create a URL for feedback on a particular page in a project."""
@@ -48,7 +51,16 @@ def feedback_form_url(project, page):
 
 html_context['feedback_form_url'] = feedback_form_url
 
-html_context['help_url'] = 'http://open.edx.org'
+
+html_context['help_url'] = GENERIC_HELP
+
+def set_audience(tag):
+    """Used from specific conf.py files to set the audience for a book."""
+    if tag == "Open_edX":
+        html_context['help_url'] = OPEN_HELP
+    elif tag == "Partners":
+        html_context['help_url'] = PARTNER_HELP
+
 
 # General information about the project.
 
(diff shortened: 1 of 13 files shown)


nedbat:
@edx/doc Opinions?

shared/conf.py
```diff
@@ -30,14 +31,36 @@
...
+FEEDBACK_FORM_FMT = "https://docs.google.com/forms/d/1gGBVXUS84Q8jVEJdlpYYIkOQmaCg5JnzBTG1x4ASWKM/viewform?entry.930288665&entry.672882205={url}"
+OPEN_HELP = "http://open.edx.org/help"              # JUST A PLACEHOLDER
+PARTNER_HELP = "http://partners.edx.org/help"       # DON'T FREAK, JUST ANOTHER PLACEHOLDER
+GENERIC_HELP = "http://help.com"                    # OBVIOUSLY A JOKE
+
+def feedback_form_url(project, page):
+    """Create a URL for feedback on a particular page in a project."""
+    return FEEDBACK_FORM_FMT.format(url=urllib.quote("{}: {}".format(project, page)))
+
+html_context['feedback_form_url'] = feedback_form_url
```
pdesjardins, line 80: Where do the values for project, page come from? Are those always available because they're set earlier in the shared conf.py?

Oh, I see those are Sphinx global variables. So we might end up seeing the filename of the source .rst file. It might be better to build this link using Javascript in order to get the current URL of the page. Or maybe let's set up a test build and see what we capture with this. I'm wondering whether frequently used .rst filenames, like index.rst will be ambiguous within a project.

pdesjardins:
@nedbat, do you have a draft project for this on RTD? I'm curious to see how the feedback link behaves.

pdesjardins:
@nedbat I checked out your branch and built locally. Looks great. I confirmed that the page identifying information collected by the form is:

Building and Running an edX Course: index
Building and Running an edX Course: video/index
Building and Running an edX Course: exercises_tools/index

That looks like plenty of information about the context of any feedback.

I think this is excellent, thanks! I would give this a 👍  when we have the real support URLs and have a permanent feedback form in place.

nedbat:
@pdesjardins thanks.  Will you be making a permanent feedback form?  I can link in whatever form you like.

nedbat:
@edx/doc I implemented the category/audience split from this spreadsheet: https://docs.google.com/spreadsheets/d/1X0x58eWeypxRjz-kMtn7z1IatrFn_Co2hquAWU_5b48/edit#gid=0
The books that aren't in this repo will have to be handled separately.  The help link doesn't appear at all on the Open learners books.
The feedback link still goes to Peter's prototype form until we have a final one to use.

lamagnifica:
@nedbat thanks for the POC! We still have a ways to go to get the link destinations finalized though.

pdesjardins:
I filed an IT request for advice about creating a non-personal Google form for this. After we figure out where that form should be, I'll create it.

pdesjardins:
@nedbat Here's the URL for the permanent feedback form:

https://docs.google.com/forms/d/1T5QGnYb_QnQoMO7T_eatq02miPTY40WVe3cgGphNAdY/viewform?entry.1952574704&entry.241692674=include-page-id-here

Replace "include-page-id-here" with the page identifying strings you are pulling from the page.

lamagnifica:
:+1: on the form and "initial" destinations.

PR #991, opened 2016-04-22
Dogwood.3 release note

nedbat:


nedbat:
@lamagnifica simple enough, yes?

lamagnifica:
:+1: 
SuperNit says, if your list only has one item, don't use a bullet (it's just a sentence, not a list).

lamagnifica:
:smile:

PR #996, opened 2016-04-26
Fix support level for new drag and drop

catong:
@pdesjardins The previous version of D&D was only provisionally supported. I did not update the level of support in the main table in the previous PR; fixing in this small PR.


catong:
Merging https://github.com/edx/edx-documentation/pull/996

PR #997, opened 2016-04-27
Make a nicer copyright, with image, link, etc.

nedbat:


nedbat:
@lamagnifica as we hacked.

lamagnifica:
:+1:

PR #998, opened 2016-04-28
Updated import guidelines

clintonb:
Using the actual PEP 8 guideline, instead of the edX customization. The customization is rather burdensome to maintain—some tools/IDEs (e.g. PyCharm) automatically adhere to PEP 8, but cannot be configured with additional customizations—and offers no obvious value over the PEP 8 recommendation.

### DIFF
--- a/en_us/developers/source/style_guides/python_guidelines.rst
+++ b/en_us/developers/source/style_guides/python_guidelines.rst
@@ -90,11 +90,10 @@ Imports Order
 PEP 8 recommends a most-general to most-specific import order, which means this order:
 
 * Standard library imports
-* Third Party Library imports
-* Core Django imports
-* Third party Django app imports
-* Other edX repo imports
+* Third party library imports
 * Local imports
+
+Each group of imports should be alphabetized; and, you should put a blank line between each group.
 
 *******************************
 Pylint Guidelines and Practices


en_us/developers/source/style_guides/python_guidelines.rst
```diff
@@ -90,12 +90,11 @@ Imports Order
 PEP 8 recommends a most-general to most-specific import order, which means this order:
 
 * Standard library imports
-* Third Party Library imports
-* Core Django imports
-* Third party Django app imports
-* Other edX repo imports
+* Third party library imports
```
nedbat, line 93: Is there a problem with the more nuanced approach? Should edx-platform treat XBlock as a third-party library just like requests?  That seems odd to me.

clintonb, line 93: I consider any code not in the current repo to be third party, regardless of the author(s).

rlucioni:
👍

en_us/developers/source/style_guides/python_guidelines.rst
nedbat, line 93: I believe the intent of the import order is to aid in reading the code, and the "most general to most specific" groups is the important thing.  If we have a more elaborate system, with more than three groups, I think it is good to adapt the standard.
To me, it seems useful to have "other edX libraries" separated out from all the generic third-party libraries we use.
I haven't heard what the problem is with the more nuanced system.

cpennington, line 93: I prefer the more nuanced version as well. It would be nice if we could automate checking for it, although that might just require a big whitelist of core vs 3rd party django, and edx vs 3rd party libraries.

rlucioni, line 93: It's worth noting that pylint already includes a check for the import order proposed here.

cpennington, line 93: Yes, pylint can check core vs non-core libraries, which will pass in both the nuanced and broad versions of the import order (assuming it's not checking alphabetization).

clintonb, line 93: My biggest issue is that none of my tools know the difference between an edX library and other libraries. Instead of a quick key press to re-order imports, I have to manually order them. The time to do that adds up.

Many of our libraries are prefixed with `edx`. If they are alphabetized, they are already grouped. The same goes for Django. I'm missing the value of adding the extra new line between those groups.

I can't think of a time where the import ordering saved me time and/or made development easier. I can point to numerous moments where import ordering resulted in more time spent having to order imports.

nedbat, line 93: I'm more concerned with the time a reader needs to invest, than the time a writer does.  I'm not a fan of giving in to simplistic tools. :)

cpennington, line 93: Once the imports are ordered correctly (however we decide correct is), then it doesn't matter, right? When you add an import, putting it in the right place is the same cost in either case.

jimabramson:
In both internal and external PRs half the feedback I give ends up being about import organization, which I feel is time better spent on more important matters.  The nuanced distinctions should either be more precisely defined or dispensed with IMO, and dispensing with them is much simpler.  Also, treating xblocks and other edx-oriented packages as third party library imports strikes me as healthy dogfood.  👍

cpennington:
@nedbat: Thoughts on a strategy of "unless/until we can lint for it, we shouldn't require it", at least for things like this?

nedbat:
I think that we should decide on the ideal for our code, and then decide how to check and implement that ideal.  It seems silly to me to say now, "Do it like this," and then in a year when a new linter is out, we change what our ideal is.  I would rather disable the message in the linter, even centrally.  The point of these conventions is to agree on what code should look like, and then make it look that way.  Automating that is a good thing, but don't let the tail wag the dog.

If we find ourselves being too picky about import ordering, then say it's suggested, and that we shouldn't take up review time with it. We're in charge here, we can decide what we want to do, tools aside.

lamagnifica:
Hi @clintonb I noticed that this PR has not yet merged. Can the discussion in the comments be brought to a resolution? Let me  or anyone on the doc team know if we can assist in moving this forward.

catong:
Hi @clintonb Please let us know if we can help in any way, or if this PR might be closed? Thanks!

lamagnifica:
Closing until some resolution and directions can be established.

lamagnifica:
Closing until some resolution and directions can be established.

PR #1004, opened 2016-04-28
Addresses Studio URL value for XSeries videos

lamagnifica:
## [DOC-2873](https://openedx.atlassian.net/browse/DOC-2873)

Adds instructions for About videos for XSeries. I decided to make an entirely separate section for the one for About videos for edx.org courses, since we are already making a distinction about Edge vs. edx.org.
### Reviewers
- [x] Subject matter expert: @rjgrubb
- [x] Doc team review (sanity check): @pdesjardins @catong @srpearce 
- [x] PM review: @jhendersonedx
### Testing
### Post-review
- [ ] Add description to release notes task as a comment
- [x] Squash commits

### DIFF
--- a/en_us/shared/set_up_course/setting_up_student_view.rst
+++ b/en_us/shared/set_up_course/setting_up_student_view.rst
@@ -16,7 +16,7 @@ see :ref:`SFD Dashboard Settings Profile`.
 .. only:: Partners
 
   For courses running on edX Edge, you configure these course details in
-  Studio. For courses running on edX.org, you work directly with your partner
+  Studio. For courses running on edx.org, you work directly with your partner
   manager to configure some of these course details.
 
 .. contents::
@@ -26,8 +26,8 @@ see :ref:`SFD Dashboard Settings Profile`.
 For information about setting important dates for the course, see
 :ref:`Scheduling Your Course`.
 
-For information about setting the course license, see :ref:`Licensing a
-Course`.
+For information about setting the licensing for the course, see
+:ref:`Licensing a Course`.
 
 .. _The Learner Dashboard:
 
@@ -74,9 +74,9 @@ might decide to enroll based on the content of the page.
   then visible on the course **About** page. For more information, see
   :ref:`installation:Add CourseTalk`.
 
-******************************************
-Add Information to the Course About Page
-******************************************
+===========================================
+Adding Information to the Course About Page
+===========================================
 
 .. only:: Open_edX
 
@@ -85,14 +85,18 @@ Add Information to the Course About Page
 .. only:: Partners
 
   For courses running on edX Edge, you configure the contents of this page in
-  Studio, as described in this topic. For courses running on edX.org, you work
+  Studio, as described in this topic. For courses running on edx.org, you work
   directly with your partner manager to configure the contents of this page.
+
+
+.. contents::
+  :local:
+  :depth: 2
 
 .. _Describe Your Course:
 
-===========================
 Describe Your Course
-===========================
+*********************
 
 Learners see the description of your course on the course **About** page. For
 example, the course description is circled in the following course **About**
@@ -104,7 +108,7 @@ page.
 
 .. only:: Partners
 
-  .. note:: For courses running on edX.org, you must communicate the course
+  .. note:: For courses running on edx.org, you must communicate the course
    description to your edX partner manager to ensure that the information
    on the course **About** page is accurate.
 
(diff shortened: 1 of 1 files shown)


en_us/shared/set_up_course/setting_up_student_view.rst
```diff
@@ -182,8 +182,8 @@ To add a course image, follow these steps.
 Adding a Course About Video
 ===========================
 
-The course About video should excite and entice potential learners to enroll, and
-reveal some of the personality that the course team brings to the course.
+The course About video should excite and entice potential learners to enroll,
```
pdesjardins: Not part of your changes, the capitalization of "about" seems incorrect to me. I'd do a global lower-casing. Also, the comma should go, I think.

Feel free to ignore spurious proofreading.

lamagnifica: agreed. we as a team need to get on the same page about uc/lc however, as each time this page gets reviewed, I get different feedback about how to present this concept...

en_us/shared/set_up_course/setting_up_student_view.rst
```diff
@@ -272,6 +273,47 @@ course content videos.
...
+      complete, the message "File Upload Complete" appears.
+
+  .. _XSeries About Video:
+
+  Add an XSeries About Video to edx.org
+  **************************************
+
+  To add an About video for an XSeries that is running on `edx.org`_ only, follow these steps.
+
+  #. Locate the video file on your computer. For example, on a Mac,
```
pdesjardins: Mac seems a little familiar. 

For example, if you are using an Apple Macintosh OS

Or omit the second sentence.

en_us/shared/set_up_course/setting_up_student_view.rst
```diff
@@ -272,6 +273,47 @@ course content videos.
...
+
+  #. Drag the video file from the local directory into the **Drop files here to
+     upload** field. You can also click inside this field to browse to the
+     file.
+
+     The file upload process begins immediately.
+
+     .. important:: Do not close the browser tab or window, or use it to go to
+      another URL while the file is uploading. When the upload process is
+      complete, the message "File Upload Complete" appears.
```
pdesjardins, line 292: Since the message is a UI label, like a button label, maybe **File Upload Complete** instead of quotation marks?

pdesjardins:
Looks good! 👍

en_us/shared/set_up_course/setting_up_student_view.rst
lamagnifica: ditto on the contradictory guidance... @nedbat requested a change to "Mac" alone as that is how Apple is presenting its product now

en_us/shared/set_up_course/setting_up_student_view.rst
lamagnifica, line 292: It's a message, not a label for a control. I thought we bolded labels for controls, but quoted messages...

"Microsoft style
If you receive a “There is no CD disc in the drive” message, insert a new CD disc in the drive and try
again."
(p76)
"Do not use special formatting (such as monospace or bold)" (p75)

en_us/shared/set_up_course/setting_up_student_view.rst
pdesjardins: Sounds good, go with it. 

One thing I was thinking of was that I'm not sure how globally "on a Mac" will be understood. It sounds like the way Americans talk to Americans.

en_us/shared/set_up_course/setting_up_student_view.rst
pdesjardins, line 292: Got it, thanks for researching.

en_us/shared/set_up_course/setting_up_student_view.rst
catong: I agree with Peter about idiomatic use of "on a..." not being ideal for globalized English. How about "if you are using a Mac machine"? That sounds kind of strange too.

rjgrubb: @lamagnifica Thanks Alison! XSeries steps look good 👍

jhendersonedx: I like the steps. Do we need to explicitly say that the URL does not have to actually work in real life? I don't want folks to think it has to be working when really we're just adding a random identifying to an existing prefix.

en_us/shared/set_up_course/setting_up_student_view.rst
```diff
@@ -272,6 +273,47 @@ course content videos.
...
+
+     You can abbreviate the full name of the XSeries. However, the information
+     that you enter should clearly identify the XSeries.
+
+  #. In the **edX Studio Course URL** field, add identifying information
+     about the XSeries to the Studio URL that is provided. For example,
+     ``https://studio.edx.org/course/XSeries_edX_VideoX``.
+
+     The value in this field must begin with
+     ``https://studio.edx.org/course/``.
```
jhendersonedx: Something like...

The value in this field must begin with
`https://studio.edx.org/course/` but does not need to resolve to a working URL.

lamagnifica: Got it. Thanks @jhendersonedx

en_us/shared/set_up_course/setting_up_student_view.rst
lamagnifica: @pdesjardins and @catong I went with "Mac computer" per Apples trademark guidelines both here and on line 261, and added the registered trademark symbol on line 261 (doh!).

lamagnifica:
@jhendersonedx please take another look.

jhendersonedx:
👍

PR #1013, opened 2016-05-04
fix formatting of dir structure

lamagnifica:


pdesjardins:
Looks sane, if it builds as expected 👍

PR #1017, opened 2016-05-09
Fix readthedocs JSON builds

nedbat:


nedbat:
@pdesjardins I think this fixes the build problems.

pdesjardins:
Looks good, 👍

PR #1030, opened 2016-05-17
DOC-2847 Course update dates

catong:
DOC-2847. This PR updates doc for course updates to reflect changes made in TNL-4115. The date field now enforces a valid date. Edits include fixing the note position and numbering that was restarting after the nested unordered list.

REVIEWERS:
- [x] @efischer19 
- [x] @srpearce or @lamagnifica or @pdesjardins 

FYI @jaakana, @sstack22

### DIFF
--- a/en_us/shared/course_assets/handouts_updates.rst
+++ b/en_us/shared/course_assets/handouts_updates.rst
@@ -23,31 +23,43 @@ Add a Course Update
 **********************
 
 You add updates to notify learners of upcoming exams or deadlines, changes in
-the course schedule, or to make other announcements.
+the course schedule, or to make other announcements. Updates are visible to
+enrolled learners as soon as you post them.
+
+Each update requires a date for identification and sorting purposes. In the
+LMS, updates display on the **Home** page in reverse chronological order (most
+recent at the top), based on the date associated with each update.
 
 To add a course update, follow these steps.
 
 #. From the **Content** menu, select **Updates**.
 #. Select **New Update**.
-#. Enter your update in the text editor that opens.
+#. Create your update in the text editor that opens.
 
-  * This editor is like the :ref:`raw HTML editor<The Raw HTML Editor>` that is
-    provided for HTML components. To format your update, you must supply HTML
-    tags. The :ref:`visual editor<The Visual Editor>` is not provided.
+   * Enter text for the update, using HTML tags for formatting. This editor is
+     like the :ref:`raw HTML editor<The Raw HTML Editor>` that is provided for
+     HTML components. The :ref:`visual editor<The Visual Editor>` is not
+     provided.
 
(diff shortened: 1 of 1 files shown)


srpearce:
👍

en_us/shared/course_assets/handouts_updates.rst
```diff
@@ -23,31 +23,44 @@ Add a Course Update
 **********************
 
 You add updates to notify learners of upcoming exams or deadlines, changes in
-the course schedule, or to make other announcements.
+the course schedule, or to make other announcements. Updates are visible to
+enrolled learners as soon as you post them.
+
+Each update requires a date for identification and sorting purposes. On the
+**Home** page of the course, updates are displayed in reverse chronological
```
lamagnifica: Sylvia's got me thinking about active voice...
maybe start with "In the LMS, updates display on the **Home** page in reverse chronological...

en_us/shared/course_assets/handouts_updates.rst
```diff
@@ -23,31 +23,44 @@ Add a Course Update
...
+        quotation marks and apostrophes.
 
-  * This editor is like the :ref:`raw HTML editor<The Raw HTML Editor>` that is
-    provided for HTML components. To format your update, you must supply HTML
-    tags. The :ref:`visual editor<The Visual Editor>` is not provided.
+   * Specify a date for this update. By default, today's date is already
+     entered. You can change the date using the calendar tool, or by entering
+     some other valid date. This date is used only for identification and
+     sorting purposes, and does not control when the update becomes visible to
+     learners.
```
lamagnifica, line 54: I think that the "the update is visible to all enrolled learners as soon as you post it." statement would still be useful here.

lamagnifica:
2 suggestions 👍

efischer19:
This looks good to me, I don't think we need to call out the reverse-compatibility features (previously entered invalid dates are sorted as if the date is today but displayed with whatever invalid value they have, new values are forced to be valid).

👍

catong:
@efischer19  Agreed re calling out the reverse-compatibility.  The behavior you put in place for those cases is logical and not disruptive, so I thought it would be unnecessarily confusing to refer to existing cases. It's also unlikely that course teams would refer to the documentation if they have already been posting course updates.

en_us/shared/course_assets/handouts_updates.rst
catong, line 54: It feels too repetitive to say this 3 times in a short topic. I tried to put it back, but I decided that in this step it add more clarity to say that the specified date does not control visibility. In the next numbered step, we do say "Your update appears ... immediately." so I think all these mentions are sufficient in combination.

catong:
Merging https://github.com/edx/edx-documentation/pull/1030

PR #1034, opened 2016-05-19
Clarifies that OLX is the tool being used

lamagnifica:
## [DOC-2970](https://openedx.atlassian.net/browse/DOC-2970)

The descriptions in the OLX guide and in the Buliding & Running guide are very similar for creating problem components. In the OLX guide only, there is info on how to add attributes to define settings. It's quite possible that someone would open the OLX guide when they were looking for info about how to use Studio, and get confused. 
### Reviewers

Possible roles follow. PR submitter checks the boxes after each reviewer finishes and gives :+1:. 
- [ ] Subject matter expert: @muhammad-ammar 
- [ ] Subject matter expert: 
- [ ] Doc team review (sanity check): @pdesjardins @catong @srpearce 
- [ ] Product review:
- [ ] Partner support: 
- [ ] PM review: 

### Testing
### Post-review
- [ ] Add description to release notes task as a comment
- [ ] Squash commits

### DIFF
--- a/en_us/olx/source/problem-xml/create_problem.rst
+++ b/en_us/olx/source/problem-xml/create_problem.rst
@@ -61,7 +61,7 @@ quickly and accurately.
 For more information about metrics for your course's problem components, see
 `Using edX Insights`_.
 
-You set the display name as an attribute of the ``problem`` element.
+With OLX, you set the display name as an attribute of the ``problem`` element.
 
 .. code-block:: xml
 
@@ -83,7 +83,8 @@ problems defaults to that number, and cannot be set to unlimited.
  higher are included on the Learner Answer Distribution report that you can
  download during your course.
 
-You set the maximum attempts as an attribute of the ``problem`` element.
+With OLX, you set the maximum attempts as an attribute of the ``problem``
+element.
 
 .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


srpearce:
👍

muhammad-ammar:
@lamagnifica Should we mention that setting attributes on `<problem>` element are applicable only in course import/export or is this understood?

lamagnifica:
@muhammad-ammar unfortunately the OLX guide is still at version 0. It was a super high priority at one point in 2015, but then was deprioritized. My assumption has been that the need to export and then import was understood, but it looks like there is just the briefest of introductions (http://edx.readthedocs.io/projects/edx-open-learning-xml/en/latest/what-is-olx.html) but no explicit guidance about how to export a course created in Studio to edit and then reimport, or to create the files from scratch and import (I'm not sure if anyone actually does that, but it seems to be a use case that this guide was intended to address...) I would rather not spend a whole lot of time on this guide -- we know it needs more work, but it just hasn't been a priority. We do have a DOC epic for completing the OLX guide if you'd like to add a story....

muhammad-ammar:
@lamagnifica I am little bit confused and not sure what to do. You are an expert on this and i think you should do what you think is best. :)

lamagnifica:
OK, thanks @muhammad-ammar . I opened DOC-2979 to add info about using OLX to edit course files exported from Studio, which you then re-import.

PR #1042, opened 2016-05-24
Move Gen Info section of learner's guide to end

srpearce:
## [DOC-2993](https://openedx.atlassian.net/browse/DOC-2993)

This PR moves the "General Information" section of the learner's guide to the end. This information, which mentions GitHub and various other doc resources, isn't applicable to most learners and could cause confusion because it's the first thing the learner sees. Also moved the less-relevant "licensing" section to later in the doc. (NB: This doc could use a reorg! Will make a separate task for that.)
- [ ] doc review (sanity check): @lamagnifica or @pdesjardins

### DIFF
--- a/en_us/students/source/front_matter/change_log.rst
+++ b/en_us/students/source/front_matter/change_log.rst
@@ -2,27 +2,15 @@
 Change Log
 ############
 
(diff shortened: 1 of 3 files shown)


lamagnifica:
👍 💃

PR #1043, opened 2016-05-24
As a stopgap, indicate release is not live

lamagnifica:
@srpearce @pdesjardins what do you think of this as a quick fix?
I changed gh-pages back to 17 May, and the portal announcement has not gone out yet. This will at least alert people who might have been waiting on the release that something has happened....


srpearce:
👍

PR #1046, opened 2016-05-26
Sync'd the order of the common sections

lamagnifica:
@srpearce @catong @pdesjardins  can I get a quick thumb on this?  I stepped on a change that Sylvia made, and realized the two versions of the guide had diverged in terms of section order…


srpearce:
👍

PR #1024, opened 2016-05-12
Ahodges/dc guidance

lamagnifica:
## [DOC-2870](https://openedx.atlassian.net/browse/DOC-2870)

Adds a new section of guidelines for data security.
### Date Needed

17 May 2016
### Reviewers

Possible roles follow. PR submitter checks the boxes after each reviewer finishes and gives :+1:. 
- [ ] Subject matter expert: @stroilova 
- [x] Subject matter expert: Nell (by email) 
- [ ] Doc team review (**sanity check**): @pdesjardins @catong @srpearce 
- [ ] Product review:
- [ ] Partner support: 
- [ ] PM review: 

FYI: @ninabeth, @e0d 

http://draft-research-guide-reorg-2016-05-09.readthedocs.io/en/latest/data_czars/data_security.html
### Testing
- [x] Ran ./run_tests.sh without warnings or errors
### Post-review
- [ ] Add description to release notes task as a comment
- [ ] Squash commits

PR #1053, opened 2016-05-31
Sylvia's copy edit of Nell's Data Czar Security Guidance

lamagnifica:
[X] Hard copy to Nell with before and after changes marked up
[ ] Sylvia after I hear back from Nell and make any changes she requires, I’ll tag you for thumbs.

https://github.com/edx/edx-documentation/pull/1024/commits/7cd9bc9cbd6060d031eb94e5450fbf564d761517

### DIFF
--- a/en_us/data/source/data_czars/data_security.rst
+++ b/en_us/data/source/data_czars/data_security.rst
@@ -16,11 +16,11 @@ periodically update these guidelines to reflect changes in best practices.
   :local:
   :depth: 1
 
-**************************
-Who Should Be Data Czar?
-**************************
+****************************
+Who Should Be the Data Czar?
+****************************
 
-The data czar should be a person on staff who is able to fulfill these
+The data czar should be a person on staff who is able to fulfill the following
 responsibilities.
 
 * Prepared to serve as the single point of contact and accountability for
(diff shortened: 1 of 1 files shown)


lamagnifica:
Message from Nell:
HI Alison,

Sorry for taking so long to follow-up with you about Sylvia's suggested edits.

Her draft looks great!  

The only edit I'd like to undo is in Section 3.2.2 - the opening phrase originally said, "As a rule..." and Sylvia suggested, "In most cases..."

No way for you to know this, but we intentionally chose that phraseology to convey the notion of a rule - we really want this to be a hard-and-fast rule, but as a practical matter, we've come to accept that we have to allow exceptions now and again.  But we liked characterizing it as a rule (in a soft way).

Does that make sense?  If you or Sylvia has suggestions for a better articulation that addresses our "rule" concern - or if you think our original language is just too goofy/confusing/etc. - please let me know.

Otherwise, everything else looks fantastic, I'm continually impressed by what a big difference your team makes with these improvements.  Kudos!

Many thanks,
Nell

lamagnifica:
@srpearce please do a final sanity check

srpearce:
👍

PR #1058, opened 2016-06-02
DOC-3013: Modifications to Bulk Email "Send To" Groups; allow multi-select

catong:
This PR updates bulk email procedures for course teams to reflect new addressing options in bulk email, including ability to select more than one of the groups. (TNL-4356)
It does not include ability to select cohorts as a "send to" option - this change will be covered in a separate PR when the code merges.

Reviewers:
- [x] @efischer19 
- [x] @lamagnifica 

FYI @sstack22, @jaakana

### DIFF
--- a/en_us/shared/manage_live_course/bulk_email.rst
+++ b/en_us/shared/manage_live_course/bulk_email.rst
@@ -11,11 +11,11 @@ and other material. All course contributors who are assigned the Staff or
 Admin role can use bulk emails to communicate with course participants before,
 during, and after the course run.
 
-.. note::
-  Some courses use third party services such as MailChimp to send bulk email.
-  Do not use both a third party service and the edX bulk email service. If you
-  use more than one service to send email, your messages are more likely to be
-  be marked as spam, and learners might not see them.
+.. note::   Some courses use third party services such as MailChimp to send
+   bulk email. Do not use both a third party service as well as the edX bulk
+   email service. If you use more than one service to send email message, your
+   messages are more likely to be be marked as spam, and learners might not
+   see them.
 
 This section contains the following topics.
 
@@ -27,27 +27,42 @@ This section contains the following topics.
 Message Addressing
 *************************
 
-When you send an email message from the instructor dashboard, you choose its
-recipients by selecting one of these predefined groups.
+When you send a bulk email message from the instructor dashboard, you choose
+its recipients by selecting a **Send to** option. You can select one or more
+recipient groups for each message.
+
+When you send a bulk email message to more than one recipient group, duplicate
+recipients are filtered out, so that someone who belongs to more than one of
+the recipient groups (for example, a course team member who is also enrolled
+in the course) only receives one copy of the message.
 
 * **Myself**, to test out a message before sending it to a larger group.
-* **Staff and admins**, to contact other members of the course
-  team.
-* **All (learners, staff and admins)**, to communicate with currently
-  enrolled learners and the course team.
 
-  Email messages are sent to all enrolled learners with these exceptions.
+* **Staff and Administrators**, to send a message to members of the course
+  team who have staff and admin privileges. For information about course
+  team member privileges, see :ref:`Course_Staffing`.
+
+* **All Learners**, to send a message to all currently enrolled learners. This
+  recipient group also includes members of the course team if they are
+  enrolled in the course.
+
+  Email messages are sent to all enrolled learners, with these exceptions.
 
   * Learners who have not replied to the account activation email message
     that is sent when they registered on edx.org.
+
   * Learners who have opted out of receiving email messages through the **Email
     Settings** link, which is present for each course on the **Current
     Courses** dashboard.
 
-Before you use the bulk email feature, consider that messages **cannot be
-canceled** after they are sent. If you plan to send a message to all course
-participants, be sure to review each draft carefully, and send the message to
-yourself first for thorough testing.
+To send an email message to all course team members as well as all learners,
+select both **Staff and Administrators** and **All Learners**. Recipients who
+are in both groups will receive only one copy of the email message.
+
+.. Important:: You cannot cancel bulk email messages after you send them.
+   Before you send a message to a wide audience, review your draft carefully.
+   You can send the message only to yourself to make sure it displays as you
+   intend it to, and that any links or media function correctly.
 
 .. _Options for Email Message Text:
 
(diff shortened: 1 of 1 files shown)


en_us/shared/manage_live_course/bulk_email.rst
```diff
@@ -27,14 +27,13 @@ This section contains the following topics.
 Message Addressing
 *************************
 
-When you send an email message from the instructor dashboard, you choose its
-recipients by selecting one of these predefined groups.
+When you send an email message from the Instructor Dashboard, you choose its
```
lamagnifica: lc instructor dashboard

en_us/shared/manage_live_course/bulk_email.rst
```diff
@@ -27,14 +27,13 @@ This section contains the following topics.
...
+recipients by selecting a **Send to** option. If applicable, you can select
```
lamagnifica: maybe just, "You can select one or more of the following recipients for each message"

en_us/shared/manage_live_course/bulk_email.rst
```diff
@@ -27,14 +27,13 @@ This section contains the following topics.
...
+more than one option at a time.
 
 * **Myself**, to test out a message before sending it to a larger group.
-* **Staff and admins**, to contact other members of the course
-  team.
-* **All (learners, staff and admins)**, to communicate with currently
-  enrolled learners and the course team.
+* **Staff and admins**, to contact other members of the course team.
+* **All learners**, to communicate with all currently enrolled learners.
```
lamagnifica: does this still include staff and admins? or does it exclude staff and admins, but now you can check both check boxes if you want to send to them both?

en_us/shared/manage_live_course/bulk_email.rst
```diff
@@ -44,10 +43,10 @@ recipients by selecting one of these predefined groups.
     Settings** link, which is present for each course on the **Current
     Courses** dashboard.
 
-Before you use the bulk email feature, consider that messages **cannot be
-canceled** after they are sent. If you plan to send a message to all course
-participants, be sure to review each draft carefully, and send the message to
-yourself first for thorough testing.
+.. note:: Before you use the bulk email feature, consider that messages cannot
```
lamagnifica: maybe even make this an "important"?

en_us/shared/manage_live_course/bulk_email.rst
```diff
@@ -44,10 +43,10 @@ recipients by selecting one of these predefined groups.
...
+   be cancelled after they are sent. Before sending a message, be sure to
+   review each draft carefully, and send the message to yourself first for
```
lamagnifica: ^only to yourself? (meaning, just select Myself and no other possible recipients at least once...)

en_us/shared/manage_live_course/bulk_email.rst
```diff
@@ -111,12 +110,16 @@ To send an email message to course participants, follow these steps.
 
 #. Select **Instructor**, and then select **Email**.
 
-#. Select who you want to send the message to from the **Send to** drop-down
-   list. You select from the following options.
+#. Select email recipients from the **Send to** list. If applicable, you can
```
lamagnifica: list --> options

lamagnifica:
question for @efischer19 about what "all learners" means now. 
Otherwise, a few suggestions to use as make sense.
👍

en_us/shared/manage_live_course/bulk_email.rst
lamagnifica: (I think it's worth being clear in the doc if "All learners" means everyone who is enrolled, which includes anyone who is accessing the course in the LMS like admins and staff...)

efischer19: As I understand it, it is possible to be a staff member and _not_ be enrolled in the course, which is what the `Staff and admins` box is for. The `All learners` options is "everyone who is enrolled in the course" which may (but doesn't have to) include instructors.

To ensure sending to `All` (as the previous option existed), you'd need to select both options. Duplicates are filtered out, so a staff user who is also enrolled would still only receive the message once.

efischer19: Maybe something like:

**Staff and admin**, to contact other members of the course team whether they are enrolled or not.
**All learners**, to communicate with all currently enrolled learners.

And add something about duplicate filtering?

en_us/shared/manage_live_course/bulk_email.rst
```diff
@@ -111,12 +110,16 @@ To send an email message to course participants, follow these steps.
...
+   select more than one option.
+
+  * **Myself**
+  * **Staff and admins**
```
efischer19: **Staff and Admin**, and **All Learners**, if we're matching the screenshots exactly.

Should I change this in code, to keep plurality consistent between `Staff and admin` and `All learners`?

en_us/shared/manage_live_course/bulk_email.rst
catong: Ack: a merge conflict resolution loss!

en_us/shared/manage_live_course/bulk_email.rst
catong: @efischer19 That would be ideal - thanks! If it's not too long, "Staff and Administrators" would be best, otherwise "Staff and Admins".

en_us/shared/manage_live_course/bulk_email.rst
catong: I had already modified the wording to point out that you can select multiple options only if it makes sense - e.g. if you select a specific cohort you can't also select "all learners", but then removed the cohort selection piece from this PR. I'll change the wording so that it's oriented more towards selecting recipients than selecting options.

catong:
@lamagnifica and @efischer19 Thanks for your input. I've addressed Alison's comments and updated the label strings per Eric's comments (I used "Staff and Administrators" for now; I can update if the final string is not that).

en_us/shared/manage_live_course/bulk_email.rst
lamagnifica: "Admin" is how the role name appears on the LMS membership page and Studio Course Team page....

efischer19: I was going to suggest `Admin` over `Administrator` as well, for spacing reasons. Is the pluralization (`Admins`) still desired?

screen shot 2016-06-03 at 2 07 23 pm

catong:
@efischer19 @lamagnifica A further note about the role names. In the platform we often conflate the names of the roles with nouns describing people with the roles. In past work I've tried to make the difference more clear in doc and UI, plus for translation and globalization reasons, I don't like "Admins". IMO I don't think it's confusing to use "Staff and Administrators" in the Send to field even if our actual role name is "Admin", but I'm open to suggestions otherwise.

en_us/shared/manage_live_course/bulk_email.rst
catong: @efischer19 Our notes crossed (see my comment about role names vs descriptive noun for people with those roles in the main conversation page). To be truly parallel I still think it should be "Administrators" but if space does not allow, "Admins" (yes please to the plural) is fine.

en_us/shared/manage_live_course/bulk_email.rst
efischer19: This comment still hasn't been addressed

efischer19:
👍 from me, if the "list --> options" comment from above was ignored deliberately. Looks great!

catong:
@efischer19 Nope it wasn't deliberate - thanks for the catch!

catong:
Merging https://github.com/edx/edx-documentation/pull/1058

PR #1063, opened 2016-06-07
Delete .pr_cover_letter.md.j2

lamagnifica:
@cpennington this deletes the unused template file.
@srpearce @catong @pdesjardins FYI


pdesjardins:
👍

PR #1080, opened 2016-06-14
Adds advice to get plain text transcripts

lamagnifica:
## [DOC-3045](https://openedx.atlassian.net/browse/DOC-3045)

Due to AC-454 rework. transcripts will now be stripped of markup and rendered as plain text only. Clayton has notified the transcript service (for partners) that we recommend that they should provide transcripts in plain text only. Add a note to this effect for course teams to inform other transcript services.
### Date Needed

20 June 2016
### Reviewers

- [x] Subject matter expert: @clrux 
- [ ] Subject matter expert: 
- [x] Doc team review (copy edit): @catong @srpearce @pdesjardins 
- [ ] Product review: 
- [ ] Partner support: 
- [ ] PM review: 

FYI: @cptvitamin 
### Testing
### Post-review
- [x] Add a comment with the description of this change or link this PR to the next release notes task.
- [x] Squash commits

### DIFF
--- a/en_us/course_authors/source/video/video_course.rst
+++ b/en_us/course_authors/source/video/video_course.rst
@@ -122,9 +122,16 @@ Create or Obtain a Transcript
 ======================================
 
 To create or obtain a transcript in .srt format, you can work with a company
-that provides captioning services. To ensure quality and accuracy of
-transcripts, edX works with `3Play Media`_. To request a 3Play account at
-edX's discounted rate, contact your edX partner manager.
+that provides captioning services.
+
+The edX video player renders transcript files as plain text. EdX recommends
+that you create or obtain transcript files that use plain text only, and that
+do not include HTML markup, character entities such as ``&nbsp;``, or styling
+such as boldface or italics.
+
+To ensure quality and accuracy of transcripts, edX works with `3Play Media`_.
+To request a 3Play account at edX's discounted rate, contact your edX partner
+manager.
 
 ===================================
 Associate a Transcript with a Video
--- a/en_us/shared/course_components/create_preroll_video.rst
+++ b/en_us/shared/course_components/create_preroll_video.rst
@@ -95,10 +95,15 @@ an identifying video ID. For more information, see `Upload Video Files`_ in
 the *Processing Video Files* guide.
 
 Prepare Pre-Roll Video Transcript Files
-**********************************************************
+*****************************************
 
-You must provide at least one transcript file for your pre-roll video. You
-can provide additional transcript files in multiple languages.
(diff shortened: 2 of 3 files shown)


en_us/course_authors/source/video/video_course.rst
```diff
@@ -122,9 +122,15 @@ Create or Obtain a Transcript
 ======================================
 
 To create or obtain a transcript in .srt format, you can work with a company
-that provides captioning services. To ensure quality and accuracy of
-transcripts, edX works with `3Play Media`_. To request a 3Play account at
-edX's discounted rate, contact your edX partner manager.
+that provides captioning services.
+
+The edX video player renders transcript files as plain text. Transcripts should
```
catong: Make more direct? How about: "Create transcripts using plain text only. Do not include HTML markup, or styles such as boldface or italics."

en_us/shared/course_components/create_preroll_video.rst
```diff
@@ -95,10 +95,14 @@ an identifying video ID. For more information, see `Upload Video Files`_ in
...
 Prepare Pre-Roll Video Transcript Files
-**********************************************************
+*****************************************
 
-You must provide at least one transcript file for your pre-roll video. You
-can provide additional transcript files in multiple languages.
+You must provide at least one transcript file for your pre-roll video. The edX
+video player renders transcript files as plain text. Transcripts should be
+created using plain text only. They should not include styling such as boldface
+or italic font or HTML markup.
```
catong: See suggested wording above. The 2nd sentence should be rewritten, at least - currently it reads as if HTML markup is an example of styling.

en_us/course_authors/source/video/video_course.rst
lamagnifica: cool. I'll use the original as the release note fodder...

downzer0:
We should include language that indicates not to escape characters or use entities (i.e., use & instead of &amp;, " instead of &quot;, etc.). Maybe that's included in the "plain text" bit? Do you think we should explicitly state it?

Otherwise 👍

lamagnifica:
@clrux I'm not following your examples (could markdown have changed them on you?). I thought about including an HTML reserved character example, like an ampersand, but ended up thinking that "use plain text" would be more broadly understood than "do not use character entities".

downzer0:
Ha, yep, looks like Github made the conversions.

& instead of &.amp;
" instead of &.quot;
Etc.

lamagnifica:
@catong I think this will be in this week's release if you want to take another pass through.

catong:
Looks good 👍

PR #1093, opened 2016-06-17
Add instructions for running smaller sets of e-commerce tests

JoyceZhu:
Documents commands to run only Python tests, only Python unit tests in a specific file, or only JavaScript tests in e-commerce. The current e-commerce docs currently only describe how to run the full suite of unit tests or quality checks. 
### Reviewers
- [x] Subject matter expert: @douglashall 
- [x] Doc team review (sanity check/copy edit/dev edit):  @pdesjardins 
### Testing

### DIFF
--- a/en_us/install_operations/source/ecommerce/test_ecommerce.rst
+++ b/en_us/install_operations/source/ecommerce/test_ecommerce.rst
@@ -60,6 +60,40 @@ When you create E-Commerce tests, use the ``TestCase`` class in
 ``Partner`` objects configured. This will help you test any code that relies on
 these models, which are used for multi-tenancy.
 
+* To run all Python unit tests and quality checks, run the following command.
+
+  .. code-block:: bash
+
+      $ make validate_python
+
+* To run all Python unit tests and quality checks *in parallel*, run the
+  following command.
+
+  .. code-block:: bash
+
+      $ make fast_validate_python
+
+* To run the Python unit tests in a specific file, such as
+  ``ecommerce/courses/tests/test_utils.py``, run the following command and
+  substitute the desired file path.
+
+  .. code-block:: bash
+
+      $ DISABLE_ACCEPTANCE_TESTS=True ./manage.py test
+      ecommerce.courses.tests.test_utils --settings=ecommerce.settings.test
+      --with-ignore-docstrings --logging-level=DEBUG
+
+  Setting the DISABLE_MIGRATIONS variable significantly decreases the time
+  needed to run tests by creating the test database directly from Django model
+  definitions as opposed to applying the defined migrations.
+
+  .. code-block:: bash
+
+      $ DISABLE_MIGRATIONS=1 DISABLE_ACCEPTANCE_TESTS=True ./manage.py test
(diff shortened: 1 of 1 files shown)


en_us/install_operations/source/ecommerce/test_ecommerce.rst
```diff
@@ -60,6 +60,32 @@ When you create E-Commerce tests, use the ``TestCase`` class in
...
+  substitute the desired file path.
+
+  .. code-block:: bash
+
+      $ DISABLE_ACCEPTANCE_TESTS=True ./manage.py test
+      ecommerce.courses.tests.test_utils --settings=ecommerce.settings.test
+      --with-ignore-docstrings --logging-level=DEBUG
+
+  Adding the ``DISABLE_MIGRATIONS=1`` option will decrease the time needed to
+  run tests at the cost of skipping migrations.
```
douglashall: I would rephrase this to something like "Setting the DISABLE_MIGRATIONS=1 variable will significantly decrease the time needed to run the tests by creating the test database directly from Django model definitions as opposed to using the defined migrations." And then follow it with a code block example.

douglashall:
One comment. Otherwise looks good! 👍

en_us/install_operations/source/ecommerce/test_ecommerce.rst
```diff
@@ -60,6 +60,32 @@ When you create E-Commerce tests, use the ``TestCase`` class in
 ``Partner`` objects configured. This will help you test any code that relies on
 these models, which are used for multi-tenancy.
 
+* To run all Python unit tests and quality checks, run the following command.
```
pdesjardins, line 63: On the doc team, we wrap lines with newlines at 80 characters. We use a Sublime text editor function to do that (https://openedx.atlassian.net/wiki/display/DOC/Documentation+Tools#DocumentationTools-AddingPackagestoSublimeText). So we don't see long lines like this one. Minor point, but that's our doc rule.

en_us/install_operations/source/ecommerce/test_ecommerce.rst
```diff
@@ -60,6 +60,32 @@ When you create E-Commerce tests, use the ``TestCase`` class in
...
+
+  .. code-block:: bash
+
+      $ make validate_python
+
+* To run all Python unit tests and quality checks **in parallel**, run the
```
pdesjardins: We typically don't emphasize using bold or italics. I'm not sure I agree with that rule. Maybe just change these to single asterisks for italics?

pdesjardins:
I added two formatting comments. 👍

en_us/install_operations/source/ecommerce/test_ecommerce.rst
JoyceZhu, line 63: Thanks for the info! FYI I just tried wrapping this line with the plugin, which says it's 78 characters -- does it look too long on GitHub?

pdesjardins:
@JoyceZhu maybe it's just the display on my screen. I'm seeing wrapping. Don't worry about it though. 👍 👍 👍

PR #1105, opened 2016-06-24
Add information about third-party authentication without transmitting PII

pdesjardins:
## [DOC-2807](https://openedx.atlassian.net/browse/DOC-2807)

This PR adds information about configuring third-party authentication without transmitting PII to an edX site.

The PR also updates the highlighting language of some code-block examples, which were generating build errors in the latest version of the Sphinx publishing tool.
### Reviewers

- [ ] Subject matter expert: @bradenmacdonald 
- [ ] Subject matter expert: @smagoun 
- [ ] Doc team review (sanity check/copy edit/dev edit): @lamagnifica @catong @srpearce 
- [ ] PM review: @ksvensson

### Testing
### HTML Version (optional)
- [x] http://draft-pii-api.readthedocs.io/en/latest/configuration/tpa/eliminating_pii.html
### Post-review
- [ ] Squash commits


en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,131 @@
...
+learners to the learner records that your institution maintains, you can use
+the third-party authentication ID mapping REST API to retrieve the user ID SAML
+attribute and matching edX username for a learner. For more information about
+grade reports, see :ref:`opencoursestaff:Access_grades`.
+
+The following diagram shows how an institution that uses third-party
+authentication can match non-personally identifying edX learner usernames with
+the records that the institution holds for those learners.
+
+.. image:: ../../../../shared/images/tpa-institution-associate-edx-id-with-personal-id.png
```
bradenmacdonald, line 78: Why not just embed the .svg file directly, and get rid of the .png version? The SVG will be higher quality.

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,131 @@
...
+identifier serves as a link between the personal identifying information that
+your institution maintains for a learner and the edX profile for that user. The
+identifying string does not need to include any PII for the learner. The
+identifying string is sometimes referred to as an opaque identifier because it
+does not make the identity of the learner visible.
+
+When learners use third-party authentication to log into an edX site for the
+first time, they create an edX learner profile. The new edX learner profile is
+permanently associated with the identifying string included in the IdP
+authentication token. Learners may enter an edX username, email address, and
```
bradenmacdonald: "Learners must enter" because it is a requirement to enter a username and email address in order to use edX.

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,131 @@
...
+
+###############################################
+Eliminating PII From Third-Party Authentication
+###############################################
+
+Open edX sites and edX Edge do not require any personally identifying
+information (PII) about learners during third-party authentication. Your
+identity provider (IdP) service can send only non-personal identifiers to
+create edX site user profiles for learners. If you configure third-party
+authentication in this way, the edX site never holds any PII for your learners.
```
bradenmacdonald: I don't think we can say "the edX site never holds any PII for your learners". All we can say is that if you configure third party authentication in this way, the edX site does not require the IdP to provide PII. (Because the platform still asks for an email address, full name, and username, the user may voluntarily provide that PII, and then the edX site would still be holding PII for some of that institution's learners.)

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,131 @@
...
+The following diagram shows how an IdP can direct a learner to an edX site to
+create a learner profile, without transmitting any PII.
+
+.. image:: ../../../../shared/images/tpa-idp-create-non-personal-account.png
+  :width: 900
+  :alt: A diagram showing how an identity provider (IdP) service can create a
+      learner profile on an edX site without transmitting any personally
+      identifying information (PII).
+
+You can download a report containing grades for all learners in the edX courses
```
bradenmacdonald, line 62: Note: I think the grade report is just one example of this sort of workflow, and there may be others.

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
bradenmacdonald, line 78: In this diagram, I don't think there should be a [dotted] line directly between "course score for edX username: ..." and "Institution grade records". I think the line should be by way of the REST API client, since it provides the info that you need to map the score to the grade record.

---

_The following is optional:_

If you don't mind making the diagram more complex, you may also want to add the identity service (IdP) back in, since the "non personal identifier" would typically only be stored by the IdP and would not be part of the institution grade records.

So the full workflow would be: course grade report provides a username (e.g. MyEdxAlias) > REST API client converts username to a non-personal identifier > IdP converts non-personal identifier to a Identifier (e.g. 5391395224 in the first image) > Identifier is used to store the grade into the institution's grade records.

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,131 @@
...
+Open edX sites and edX Edge do not require any personally identifying
+information (PII) about learners during third-party authentication. Your
+identity provider (IdP) service can send only non-personal identifiers to
+create edX site user profiles for learners. If you configure third-party
+authentication in this way, the edX site never holds any PII for your learners.
+
+After you have configured third-party authentication at an edX site, the edX
+site log in page displays a link to your IdP service. Learners follow that
+link, log in using their authentication credentials for your institution, and the IdP
+directs them back to the edX log in page, along with an authentication token
```
bradenmacdonald: It would be more accurate to use the term "SAML attributes" or "SAML assertion" instead of "authentication token"

(Or say "authentication token (SAML attribute assertion statement)")

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
pdesjardins, line 78: About SVG vs PNG: I actually never thought of embedding the SVG because I was worried about browser incompatibility. I tried it and found that all the text disappeared in Chrome. I think that Inkscape might not produce the most consumable SVG code.

en_us/shared/images/tpa-idp-create-non-personal-account.svg
```diff
@@ -0,0 +1,457 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
lamagnifica, line 1: @pdesjardins this is a nit, but if the tool you're using easily supports layers, the text should go in a separate layer (looking ahead to translation). Boxes could be a bit bigger to allow for ~25% expansion as well.

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
...
+institution.
+
+After you have configured third-party authentication at an edX site, the edX
+site log in page displays a link to your IdP service. Learners follow that
+link, log in using their authentication credentials for your institution, and
+the IdP directs them back to the edX log in page, along with an authentication
+token (a SAML attribute assertion statement) containing information about the learner.
+
+The IdP authentication token includes an identifying string for a learner. This
+identifier serves as a link between the personal identifying information that
```
lamagnifica: possibly rephrase "personal identifying information" to something that isn't so similar to "personally identifiable information"? maybe just drop "personal" here? or only use "information"?

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
...
+your institution maintains for a learner and the edX profile for that user. The
```
lamagnifica: learner / user  might be confusing
profile might be confusing 
maybe, "...maintains for a user and the edX account for that user" ?

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
...
+Eliminating PII From Third-Party Authentication
+###############################################
+
+Open edX sites and edX Edge do not require any personally identifying
+information (PII) about learners during third-party authentication. Your
+identity provider (IdP) service can send only non-personal identifiers to
+create edX site user profiles for learners. If you configure third-party
+authentication in this way, the edX site never receives PII from your
+institution.
+
```
lamagnifica, line 14: add contents directive, and maybe add an h2 for the conceptual info that follows?

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
...
+The IdP authentication token includes an identifying string for a learner. This
+identifier serves as a link between the personal identifying information that
+your institution maintains for a learner and the edX profile for that user. The
+identifying string does not need to include any PII for the learner. The
+identifying string is sometimes referred to as an opaque identifier because it
+does not make the identity of the learner visible.
+
+When learners use third-party authentication to log into an edX site for the
+first time, they create an edX learner profile. The new edX learner profile is
```
lamagnifica: agreement issue, also profile vs. account again (throughout -- let's discuss if I'm not understanding a difference you're deliberately making). 
When a learner... the edX site creates a learner account

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
...
+permanently associated with the identifying string included in the IdP
+authentication token. Learners must enter an edX username, email address, and
```
lamagnifica: I know that this information is required to create the account, but is it true that learners must enter it? I thought that all required info could be passed from the other site.

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
...
+other information in their profiles. The information that learners enter does
+not need to be personally identifying. For more information, see
```
lamagnifica: if the issue is that they do not have to enter PII, maybe use that instead of "personally identifying"? 
(I think there's a lot of potential for confusion about what is meant)

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
...
+
+To avoid storing PII on an edX site, learners can enter non-personally
+identifying information in their profiles. For example, a learner can use an
+alias as a username and create a non-identifying email address to receive
+course updates.
+
+You can include information that will be included in the edX learner profile in
+the authentication token that your IdP sends during third-party authentication.
+Authentication tokens can contain a username, first name, last name, and
+username that will appear in the form fields that learners use to create an edX
```
lamagnifica: username 2x -- email address?

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
...
+identifies the learner in course discussions and in records such as course
+grades. The edX site sends course updates to the email address in a learner
+profile.
+
+To avoid storing PII on an edX site, learners can enter non-personally
+identifying information in their profiles. For example, a learner can use an
+alias as a username and create a non-identifying email address to receive
+course updates.
+
+You can include information that will be included in the edX learner profile in
```
lamagnifica: to avoid the include-include, maybe drop the edX side in this first sentence? something like
You can configure the info for the authentication token to send during third party auth. Auth tokens can contain some or all of the information that is required to create an edX account, which populates the account registration form...

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
...
+the authentication token that your IdP sends during third-party authentication.
+Authentication tokens can contain a username, first name, last name, and
+username that will appear in the form fields that learners use to create an edX
+learner profile.
+
+If you want to eliminate PII from the edX learner profiles that use third-party
+authentication, you should not include personally identifying information in
+the authentication token. The only piece of information that is required in the
```
lamagnifica: maybe add, as in the previous example, something about instead you can supply an alias as the username and a non-identifying email address

lamagnifica: or, are you saying, you should be sure to configure the auth token so that it does _not_ send this type of info, and instead leave it to the learner to fill in themselves?
If the audience = me, I'm a little confused about what I'm supposed to actually do

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
...
+authentication.
+
+You can include an email address in the third-party authentication token that
+your IdP sends when it redirects learners to an edX site. The email address
+will appear in the email address field of the edX registration page when a
+learner visits the edX site for the first time.
+
+If you supply a non-identifying email address in the authentication token, and
+make that email address an alias for a learner's institutional email address,
+the learner will receive edX course information at the instituional email
```
lamagnifica: sp institutional

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
...
+address.
+
+For example, if a learner's institutional email address is jsmith@school.edu,
+you might generate a non-identifying email address such as
+650EAB0C-E750-4F5D-9426-74F5D078A220@school.edu and make that generated address
+an alias of jsmith@school.edu. If you supply the non-identifying email address
+in the third-party authentication token, the learner can receive course email
+conveniently, without storing a personally identifying email address at the edX
+site.
+
```
lamagnifica: is there similar advice about the alias/username? (prob not, just wondering)

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
...
+learner visits the edX site for the first time.
+
+If you supply a non-identifying email address in the authentication token, and
+make that email address an alias for a learner's institutional email address,
+the learner will receive edX course information at the instituional email
+address.
+
+For example, if a learner's institutional email address is jsmith@school.edu,
+you might generate a non-identifying email address such as
+650EAB0C-E750-4F5D-9426-74F5D078A220@school.edu and make that generated address
```
lamagnifica: is this a realistic example? cause I'd want to change it if it showed up in my registration form, too!

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
...
+an alias of jsmith@school.edu. If you supply the non-identifying email address
+in the third-party authentication token, the learner can receive course email
+conveniently, without storing a personally identifying email address at the edX
+site.
+
+For more information about including an email address in a third-party
+authentication token, see :ref:`Configuration Options for SAML Providers`.
+
+.. note::
+    Learners can alter the information in the fields of the edX site registration page, before they create their profiles. If you supply an alias email address, make sure that learners understand that they should not alter it before creating their profiles.
```
lamagnifica: hmm maybe this needs to come first? 
regardless of what you configure the auth token to send, learners can change whatever is populated in the form. 
For example, you might generate an alias for the username that the learner can change to a more familiar non identifying username. However, if you generate a non-identifying email address for the learner and make that generated address an alias for the address the learner typically uses, you should be sure to caution users not to change the supplied email address.

lamagnifica:
@pdesjardins I think this will help a lot. Some potentially confusing/overloaded terms noted. On a dev edit level, I'm wondering if framing this concept more generally, in a section about how you can configure what gets sent in the auth token, with just one of the topics about how you can do it so that no PII gets transferred / stored at edX, might work. Possibly you already have a second phase in mind for that type of expansion. Or, I could definitely see the interest in configuring the token to be primarily around the PII issue, so putting those words in the top level heading makes sense from that PoV...

en_us/shared/images/tpa-idp-create-non-personal-account.svg
pdesjardins, line 1: @lamagnifica, good point about the layers. I think you're right and I'd prefer to make these easy to translate. I don't know how to do this in Inkscape in the right way at the moment. I just posted a question in the Inkscape forum. I'd like to wait on making this improvement, and then refactor this diagram and the others I've added. Sound OK?

Also, I increased the box sizes, which was bothering me also. Just needed to rethink the arrangement to accommodate.

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
```diff
@@ -0,0 +1,135 @@
+.. _eliminating_pii_third_party_authentication:
+
+###############################################
+Eliminating PII From Third-Party Authentication
+###############################################
+
+Open edX sites and edX Edge do not require any personally identifying
+information (PII) about learners during third-party authentication. Your
```
smagoun: Is PII defined someplace? We should be very clear about what we mean by PII in this case.

smagoun:
+1 from me once we define PII (or link to a definition)

@feanil you may want to review as well

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
pdesjardins: Thanks @smagoun. I added a sentence to clarify:

Open edX sites and edX Edge do not require any personally identifiable
information (PII) about learners during third-party authentication. PII is
information that can be used to reveal an individual's identity, such as a
name, identification number, or email address.

Please let me know if that's not what you meant, or if that explanation doesn't sound right.

lamagnifica: We should be careful here, as there is a legal (US) definition and other more stringent legal requirements in other parts of the world.

smagoun: +1

pdesjardins: Agreed, we should be careful. We could stay far away from trying to define PII and give "examples of PII." Do you think we need help from the legal team here? 

I'm inclined to make it clear to readers that this technical product documentation is not meant as legal guidance and that any interpretation of what constitutes PII should come from legal experts in the context the product will be used in. I could add a note stating that.

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
feanil, line 78: It looks like two of the 3 svgs you checkin generate just fine it's just `tpa-idp-create-non-personal-account.svg` which is having issues rendering text.

feanil:
:+1:

en_us/install_operations/source/configuration/tpa/eliminating_pii.rst
lamagnifica: I like your inclination! perhaps crafting a note that we can set up as an include, and reuse, would be a worthile exercise
Nell recently wrote up these data security guidelines: http://edx.readthedocs.io/projects/devdata/en/latest/data_czars/data_security.html and I noticed at the time that she does not use, or define, PII. It may be well worth a few minutes to drop her an email to ask for her guidance, at least to surface that we are adding this topic to the doc.

pdesjardins:
@lamagnifica I revised this section based on your development edit, thanks for the close read! Would you take another look?

@smagoun and @lamagnifica, I added a note clarifying that this technical product document does not provide legal guidance for handling PII (http://draft-pii-api.readthedocs.io/en/latest/configuration/tpa/eliminating_pii.html#eliminating-pii-from-third-party-authentication). I will get some edX legal advice about that. Does this seem like the right direction to you?

lamagnifica:
@pdesjardins this is great, thanks for synthesizing my feedback into your vision. 👍

pdesjardins:
Great! Thanks for the second look, @lamagnifica. We're waiting for a review of the PII description from the legal team and this can merge. I think @smagoun 's official thumb depends on that legal review.

PR #1114, opened 2016-07-06
Fixes DOC-3060

lamagnifica:
## [DOC-3060](https://openedx.atlassian.net/browse/DOC-3060)

FEDX-198 enforced a default display name for all components added to a course.
### Date Needed

This released in June
### Reviewers

- [x] Subject matter expert: @clrux 
- [ ] Subject matter expert: 
- [x] Doc team review (sanity check): @srpearce  
- [ ] Partner support: 
- [ ] PM review: 

FYI: @mmacfarlane
### Testing
### Post-review
- [x] Squash commits

### DIFF
--- a/en_us/olx/source/components/discussion-components.rst
+++ b/en_us/olx/source/components/discussion-components.rst
@@ -70,8 +70,9 @@ The ``discussion`` element contains no children.
    * - Attribute
      - Meaning
    * - ``display_name``
-     - The value that is displayed to students as the name of the discussion
-       component.
+     - Required. The value that is displayed to students as the name of the
+       discussion component. If you do not supply a ``display_name`` value,
+       "discussion" is supplied for you.
    * - ``discussion_category``
      - The name of the category for the inline discussion as shown in the main
        **Discussion** tab of the course.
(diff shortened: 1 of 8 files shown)


en_us/olx/source/components/discussion-components.rst
```diff
@@ -70,8 +70,9 @@ The ``discussion`` element contains no children.
    * - Attribute
      - Meaning
    * - ``display_name``
-     - The value that is displayed to students as the name of the discussion
-       component.
+     - Required. The value that is displayed to students as the name of the
+       discussion component. If you do not supply a ``display_name``,
```
srpearce: I would add "value" after "a `display_name`" (all instances), or change it to "a supply a value for `display_name`" - to me, names of fields/attributes/etc. function as adjectives so they don't work as standalone nouns. That might just be me, though...

srpearce:
One comment; 👍

downzer0:
You mention "required", but then suggest that it's not required, because the system will provide one.

lamagnifica:
@clrux What wording would you suggest instead? 
(to me, if you do not provide a name (or deliberately leave it out), that the system gives you one anyway means that it's required, not optional.)

downzer0:
With a grain of salt... I feel that "required" means that they _must_ add something (it's empty otherwise). We should encourage them to _update_ what's there and be more descriptive, but because there will be a value there regardless, saying a value is required feels moot.

downzer0:
After talking in person, this makes sense. Thanks for the clarification! 👍

sstack22:
👍

PR #1121, opened 2016-07-15
import safe filter before use message

alisan617:
It will be better to include an import the safe filter before using it, to avoid confusion and misstep.
### Reviewer
- [ ] @robrap 
- [ ] @catong 
### Post-review
- [ ] Squash commits

### DIFF
--- a/en_us/developers/source/conventions/safe_templates.rst
+++ b/en_us/developers/source/conventions/safe_templates.rst
@@ -1218,9 +1218,17 @@ context, so you must use one of the following safe filters.
 
 .. code-block:: mako
 
-    ## DO this
+    <%!
+    from openedx.core.djangolib.js_utils import dump_js_escaped_json
+    %>
+    ...
     ${x | n, dump_js_escaped_json}
+
     ## or
+    <%!
+    from openedx.core.djangolib.js_utils import js_escaped_string
+    %>
+    ...
     ${x | n, js_escaped_string}
 
     ## or DO this sparingly


en_us/developers/source/conventions/safe_templates.rst
```diff
@@ -1216,6 +1216,12 @@ mako-invalid-js-filter
...
 
+Be sure to import the safe filters before use.
+
+.. code-block:: mako
+
+    from openedx.core.djangolib.js_utils import js_escaped_string
+
 .. code-block:: mako
 
     ## DO this
```
robrap: Instead of a new section, I'd just follow other samples I have in the doc and include the right import directly inline:

```
    <%!
    from openedx.core.djangolib.js_utils import dump_js_escaped_json
    %>
    ...
    ${x | n, dump_js_escaped_json}

    ## or
    <%!
    from openedx.core.djangolib.js_utils import js_escaped_string
    %>
    ...
    ${x | n, js_escaped_string}

    ## or DO this sparingly
    ${x | n, decode.utf8}
```

robrap:
Thanks for setting this up @alisan617.

FYI: @catong

alisan617:
Thanks @robrap 
@catong FYI.

ready for re-review.

catong:
Alisan, I don't know if you have rebuilt the Developer's Guide to check the output, so attaching a screenshot here. This is what the output looks like. 

alisanpr

@robrap @alisan617 if that looks as intended, I'm fine with merging.

alisan617:
It looks ok to me. @robrap ?

robrap:
:+1: @alisan617: The only question is whether or not you have it in you to adjust other rule descriptions that have the exact same problem, rather than focussing on only the one that bit you.  Either way, this moves the docs in the right direction.

Thanks.

alisan617:
@robrap That is a bigger task :)  I was thinking having a note on all sections to remind people to import utilities rather than repeating the import statements in all sections. That would make the docs lean and clean.

robrap:
@alisan617: A reminder to import is another way to go about it, but would require some effort to link back to all the appropriate sections.  Anyway, we'll leave it for now and see what happens.  Thanks.

PR #1134, opened 2016-07-21
Fixed WARNINGS encountered when publishing the doc set with Sphinx 1.3.5

pdesjardins:
## [DOC-3160](https://openedx.atlassian.net/browse/DOC-3160)

Fixed WARNINGS encountered when publishing the doc set with Sphinx 1.3.5 (as Read the Docs does). All warnings were because of syntax highlighting in code examples. In some cases the fix was to correct the syntax of code examples, so Sphinx could parse them. In many cases, the fix was to change the syntax highlighting language to none.

After these changes merge, the doc team should be able to standardize our Sphinx versions on v1.3.5 an no one will see spurious warnings.
### Reviewers

- [x] Doc team review (sanity check): @catong @lamagnifica @srpearce
### Testing
### Post-review
- [ ] Squash commits

### DIFF
--- a/en_us/course_catalog_api_user_guide/source/authentication/index.rst
+++ b/en_us/course_catalog_api_user_guide/source/authentication/index.rst
@@ -193,7 +193,7 @@ To get an access token for the edX API, follow these steps.
    The following example ``access_token`` value includes an access token
    string.
 
-   .. code-block:: json
+   .. code-block:: none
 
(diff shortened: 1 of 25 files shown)


srpearce:
This LGTM—though, admittedly, I don't really know how it works. 

I imagine that after this is merged, we can upgrade to Sphinx 1.3.5 at our leisure?

pdesjardins:
Yes, after this merges we can all upgrade to v1.3.5 and should thereafter expect no warnings.

PR #1137, opened 2016-07-22
Defines the instructor dashboard and adds how you get there

lamagnifica:
## [DOC-3164](https://openedx.atlassian.net/browse/DOC-3164)

per feedback on the bulk email page, to clue people in better as to what we mean by the instructor dashboard, I added a glossary definition and also included phrasing like "[do the thing] in the LMS by selecting **Instructor** to access the instructor dashboard."
### Reviewers

- [ ] Doc team review (sanity check): @catong @lamagnifica @pdesjardins @srpearce
### Testing
### Post-review
- [ ] Squash commits


en_us/shared/set_up_course/course_prerequisites.rst
```diff
@@ -117,11 +117,10 @@ Adjust Scores in the Entrance Exam
...
 If you find an error in the exam after learners have taken it, and corrections
 to the exam are unavoidable, you have several options to rescore the exam for
-individual learners. These options are available on the Instructor Dashboard.
+individual learners. These options are available on the instructor dashboard.
 
-On the Instructor Dashboard, select **Student Admin**, and then locate the
-**Entrance Exam Grade Adjustment** section. The following options are
-available.
+In the LMS, select **Instructor** to access the instructor dashboard, and then
+select **Student Admin**. In the **Entrance Exam Grade Adjustment** section, the following options are available.
```
srpearce: Line wrap

srpearce:
Just one comment: I would orient the reader so he knows where in the LMS he should be looking for "Instructor". So, for all instances of "select[ing] **Instructor**", I would say "select[ing] the **Instructor** page."

lamagnifica:
I'm confused, @srpearce , I thought that we were in agreement that we would limit the use of specific control identifiers and rely only on the labels. It's in the notes on the edX style wiki page, and in the style guide.

srpearce:
You're referring to this, right?

> When you describe interactions with the user interface, refer to the control labels and do not describe the control types.
> 
> Incorrect: In the Randomization drop-down list, click the Never option.
> Correct: From the Randomization list, select Never.

I thought that rule was for "controls"—small items like check boxes vs. drop-down menus vs. radio buttons, which can change quickly. I don't think of "pages" and "tabs" as controls. They also aren't likely to change, and using the word "page" or "tab" provides valuable orientation.

I'm not sure we ever explicitly made that distinction, but that was my assumption for this rule.

That's my two cents, anyway.

lamagnifica:
"Remember not to refer to a selection action as "click the Courseware tab", instead, "select Courseware". " https://openedx.atlassian.net/wiki/pages/viewpage.action?pageId=22315982 April 2015

catong:
@lamagnifica @srpearce My $0.02: I'm not sure I have always remembered to do this in practice, but in theory I agree with the more minimal approach and leave out references to UI elements. Looking at this PR, I think adding the specific **Instructor** reference is a big help to readers, and is clear enough without adding "page".

srpearce:
OK, sounds good.

PR #1155, opened 2016-08-01
Adds TPA and SSO to the glossary

lamagnifica:
Per Griff, search terms single sign on and Shibboleth don't return info on third party authentication...
### Reviewers

- [ ] Subject matter expert: @griffresch 
- [ ] Subject matter expert: 
- [x] Doc team review (sanity check): @catong @pdesjardins @srpearce
- [ ] Product review:
- [ ] Partner support: 
- [ ] PM review: 

### Testing
### Post-review
- [ ] Squash commits

### DIFF
--- a/en_us/shared/glossary/glossary.rst
+++ b/en_us/shared/glossary/glossary.rst
@@ -268,9 +268,9 @@ C
 **Custom Response Problem**
 
   A custom response problem evaluates text responses from learners using an
-  embedded Python script. These problems are also called "write-your-own-
-  grader" problems. For more information, see :ref:`partnercoursestaff:Write
-  Your Own Grader`.
+  embedded Python script. These problems are also called
+  "write-your-own-grader" problems. For more information, see
+  :ref:`partnercoursestaff:Write Your Own Grader`.
 
 .. _D:
 
@@ -555,7 +555,7 @@ L
 
 **LON-CAPA**
 
-  The LearningOnline Network with Computer-Assisted Personalized Approach
+  The Learning Online Network with Computer-Assisted Personalized Approach
   e-learning platform. The structure of CAPA problem types in the edX platform
   is based on the `LON-CAPA`_ assessment system, although they are not
   compatible.
@@ -799,10 +799,19 @@ S
 
 **Simple Editor**
 
-  The graphical user interface in a problem component that contains formatting
-  buttons and is available for some problem types. For more information, see
+  The graphical user interface in a problem component that contains a toolbar
+  for adding Markdown formatting to the text you supply. The simple editor is
+  available for some problem types. For more information, see
   :ref:`partnercoursestaff:Problem Studio View`.
 
+**Single Sign-On (SSO)**
+
+  SSO is an authentication service that allows a user to access multiple
+  related applications, such as Studio and the LMS, with the same username and
+  password. The term SSO is sometimes used to refer to third party
+  authentication, which is a different type of authentication system. For
+  information about third party authentication, see
+  :ref:`Third Party Authentication<T>`.
 
 .. _Split_Test:
 
@@ -838,6 +847,18 @@ T
 
   For more information, see :ref:`partnercoursestaff:Text Input`.
 
+.. _TPA_g:
+
+**Third Party Authentication**
+
+  A system-wide configuration option that allows users who have a username and
+  password for one system, such as a campus or institutional system, to log in
+  to that system and automatically be given access to the LMS. These users do
+  not enter their system credentials in the LMS.
+
+  For more information about how system administrators can integrate an
+  instance of Open edX with a campus or institutional authentication system,
+  see :ref:`installation:Enabling Third Party Authentication`.
 
 .. _Transcript:
 


en_us/shared/glossary/glossary.rst
```diff
@@ -803,6 +803,9 @@ S
   buttons and is available for some problem types. For more information, see
   :ref:`partnercoursestaff:Problem Studio View`.
 
+**Single Sign-On (SSO)**
+
+  See :ref:`TPA_g`.
```
pdesjardins: I think we should include an explanation of what SSO is. I believe we actually use SSO in some contexts, such as Studio and the LMS (or I'm completely wrong). Maybe:

An authentication system that allows a user to access multiple services with the same user identifier and password. The term SSO is sometimes used to refer to third party authentication, which is a different type of authentication system.  For information about third-party authentication, see :ref:`TPA_g`.

en_us/shared/glossary/glossary.rst
```diff
@@ -838,6 +841,18 @@ T
 
   For more information, see :ref:`partnercoursestaff:Text Input`.
 
+.. _TPA_g:
+
+**Third Party Authentication**
+
+  A system-wide configuration option that allows users who have credentials for
+  one system, such as a campus or institutional system, to use those
```
pdesjardins: institutional system, to log into that system and automatically gain access to the LMS. User do not enter their system credentials in the LMS.

Omit the last sentence in this paragraph. It's important that the users don't enter their credentials in the edX system.

en_us/shared/glossary/glossary.rst
pdesjardins: I see this link broke in the tests. I wonder whether underscores are reserved in anchor strings. Maybe try TPA-g as the anchor?

catong: I seem to remember that hyphens definitely broke links, while underscores don't. That might have changed or I might be wrong, though.

lamagnifica: many other anchors in this file have underscores...

pdesjardins:
Re: Shibboleth, I'm able to search for it and get to TPA doc in the latest ICR guide: http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/search.html?q=shibboleth&check_keywords=yes&area=default

en_us/shared/glossary/glossary.rst
pdesjardins: So, one problem here is that our glossary terms are not headings, just bold text. So placing an anchor above **Third Party Authentication** doesn't specify the link text.

Two quick solutions:
- See :ref:`Third Party Authentication <TPA_g>`.
- See `Third Party Authentication`_. (And also change the anchor to `.. _Third Party Authentication:`)

Oddly, internal cross-references to glossary entries do seem to work when they use intersphinx and lead to the partnercoursestaff book. I still don't understand this but I'm curious!

pdesjardins: More about the link failure: I think that intersphinx includes our non-heading glossary entries as linkable objects. So I think that after this merges and we publish the guides on RTD, a link like this will work:

See :ref:`Text Input_g`.

But it will link to a randomly selected guide that contains the glossary. If you add a specific guide to the link (:ref:`partnercoursestaff:Text Input_g`) you can control which guide you link to, but since the glossary is shared, links will only work correctly in one guide.

I think glossary links should use this format:

See `Third Party Authentication`_. (And also change the anchor to .. _Third Party Authentication:)

This avoids the possibility that we change a glossary entry title and forget that we've linked to it from elsewhere in the glossary. It also avoids the problem caused by intersphinx references leading from the glossary in one guide to the glossary in another guide.

lamagnifica: aha!
an example of how to do an intra-file xref correctly in the glossary is for split test:

```
**Split Test**

  See :ref:`Content Experiment<C>`.
```

lamagnifica:
re: Shibboleth: Griff acknowledges that he probably spelled it wrong in his search

en_us/shared/glossary/glossary.rst
```diff
@@ -803,6 +803,14 @@ S
...
   :ref:`partnercoursestaff:Problem Studio View`.
 
+**Single Sign-On (SSO)**
+
+  SSO is an authentication service that allows a user to access multiple
+  related applications, such as Studio and the LMS, with the same username and
+  password. The term SSO is sometimes used to refer to third party
+  authentication, which is a different type of authentication system. For
+  information about third party authentication, see
+  :ref:`Third Party Authentication<T>`.
```
pdesjardins, line 814: I see how this will work, and I think we could improve the experience by linking directly to the entry.

My preference is `Third Party Authentication`_ and making the corresponding change (and rooting out a duplicate anchor if it's present).

But debating glossary link behavior is not part of your excellent improvement here. Let's queue this up as a separate conversation. I think we can improve glossary linking in a few other entries, also.

en_us/shared/glossary/glossary.rst
```diff
@@ -838,6 +846,18 @@ T
...
   For more information, see :ref:`partnercoursestaff:Text Input`.
 
+.. _TPA_g:
+
+**Third Party Authentication**
+
+  A system-wide configuration option that allows users who have a username and
+  password for one system, such as a campus or institutional system, to log in
+  to that system and automatically be given access to the LMS. These users do
+  not enter their system credentials in the LMS..
```
pdesjardins: Extra period here.

pdesjardins:
I noticed an extra period, and I think we should discuss glossary linking separately from this PR, otherwise, enthusiastic 👍 !

PR #1165, opened 2016-08-08
Updates event source for four problem events

lamagnifica:
## [DOC-3229](https://openedx.atlassian.net/browse/DOC-3229)

Per research by @muhammad-ammar and team, changing source of four problem-related events from server to browser.
https://openedx.atlassian.net/wiki/display/AN/CAPA+events+after+a11y+changes
### Reviewers

- [ ] Subject matter expert: @stroilova 
- [ ] Subject matter expert: @muzaffaryousaf 
- [x] Subject matter expert: @muhammad-ammar 
- [ ] Subject matter expert: @brianhw 
- [x] Doc team review (copy edit): @catong @pdesjardins 
- [ ] Partner support: 
- [ ] PM review: 

FYI: 
### Testing
### Post-review
- [ ] Squash commits

### DIFF
--- a/en_us/data/source/internal_data_formats/tracking_logs.rst
+++ b/en_us/data/source/internal_data_formats/tracking_logs.rst
@@ -25,121 +25,89 @@ Reviewing a Sample Event
 
 A sample event from an edX.log file follows. This sample was edited to remove
 personally identifiable information. Events are stored in JSON documents, which
-can be difficult to read before standard formatting is applied.
-
-.. code-block:: none
-
-    {"agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko)
-    Chrome/30.0.1599.101 Safari/537.36", "context": {"course_id": "edx/AN101/2014_T1",
-    "module": {"display_name": "Multiple Choice Questions"}, "org_id": "edx", "user_id":
-    9999999}, "event": {"answers": {"i4x-edx-AN101-problem-a0effb954cca4759994f1ac9e9434bf4_2_1":
-    "yellow", "i4x-edx-AN101-problem-a0effb954cca4759994f1ac9e9434bf4_4_1": ["choice_0", "choice_2"]},
-    "attempts": 1, "correct_map": {"i4x-edx-AN101-problem-a0effb954cca4759994f1ac9e9434bf4_2_1":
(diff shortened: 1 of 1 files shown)


en_us/data/source/internal_data_formats/tracking_logs.rst
```diff
@@ -3200,9 +3184,9 @@ successfully.
 
 .. no sample to check
 
-The server emits ``showanswer`` events when the answer to a problem is shown.
+The browser emits ``showanswer`` events when the answer to a problem is shown.
```
catong: Any way of changing this from passive to active? Are there specific triggers/user actions to answers being shown?

catong:
@lamagnifica LGTM with one question on "showanswer".

en_us/data/source/internal_data_formats/tracking_logs.rst
lamagnifica: There are, but the user triggers a series of activities, and aren't the immediate cause of the event. So if a learner select "Show Answer" the browser does a few steps, talks to the server, the server responds, and THEN emits an event/shows the answer to the learner. In this file, active voice is extremely hard to use accurately. If you think it's very important, I'll open a separate story to tackle it for the entire file, but it's likely to keep getting kicked down the street...

catong: Ah, thanks for the explanation. I'm fine with existing wording if it's deliberate. Does this mean "Reset" above is different, and the event is emitted by the user selecting "Reset"?

lamagnifica: Yeah. You could be shown an answer because you selected Show Answer, or you could be shown the answer after the final attempt is made, or ... all the settings and otpions, all the capa problems, oh my

muhammad-ammar: @lamagnifica `showanswer` event is emitted from server side. This event only contains `problem_id` in its event data and `[answers, contents]` are not included in event data.

lamagnifica: thank you @muhammad-ammar ! I completely missed that on the wiki page.

lamagnifica:
Updated showanswer with the correct info from the wiki page (thanks @muhammad-ammar ). I also realized that the example problem at the beginning of the file was an old problem_check, so I replaced it with the newly generated example.

muhammad-ammar:
@lamagnifica LGTM 👍

PR #1185, opened 2016-08-17
mattdrayer/update-insights-switch: Update command reference

mattdrayer:
@dsjen -- going through the Analytics Devstack setup process and I had to use these commands instead of what is currently described.


dsjen:
FYI @ajpal

dsjen:
👍

PR #1190, opened 2016-08-19
Send people to the central getting help page

nedbat:
We're trying to use the getting-help page as the one landing page for people needing help.

### DIFF
--- a/en_us/shared/preface.rst
+++ b/en_us/shared/preface.rst
@@ -330,14 +330,14 @@ Additional repositories are used for other projects. Our contributor agreement,
 contributor guidelines and coding conventions, and other resources are
 available in these repositories.
 
-======================
-Community Discussions
-======================
+============
+Getting Help
+============
 
-The `Community Discussions`_ page in the Open edX Portal lists different
-ways that you can ask, and answer, questions.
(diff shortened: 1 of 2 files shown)


nedbat:
@edx/doc ok?

en_us/shared/preface.rst
```diff
@@ -330,14 +330,14 @@ Additional repositories are used for other projects. Our contributor agreement,
...
 
-======================
-Community Discussions
-======================
+============
+Getting Help
+============
 
-The `Community Discussions`_ page in the Open edX Portal lists different
+The `Getting Help`_ page in the Open edX Portal lists different
```
catong, line 337: "...ask, and answer," -> "ask, and get answers to,"

"answer" made more sense when we pointed to the forum.

catong:
Looks great! One suggestion, otherwise 👍

en_us/shared/preface.rst
nedbat, line 337: Hmm, we do hope that people will start answering questions, and the Getting Help page lists the same stuff as the Discussions page, plus more.  But these sections are specifically for help seekers, so perhaps it's best to stick to that so as to avoid confusion.

PR #1192, opened 2016-08-19
Use a shorter and better download command

nedbat:


nedbat:
This command is shorter and better.  Don't you think? @gsong @cpennington @feanil

feanil:
:+1:

cpennington:
:+1:

Would it be crazy to use a link-shortener or something to make the url even easier?

feanil:
It would make it less clear what you're getting but it would make it easier for us to get analytics on how often and from where people are downloading this file.

nedbat:
I like Feanil's idea: use files with short URLs in S3.  Each file could be a script customized to install a particular configuration and version.

PR #1194, opened 2016-08-22
Add an OEP-2 compliant openedx.yaml file

cpennington:
This adds an `openedx.yaml` file, as described by OEP-2:
http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html

The data in this file was transformed from the contents of
edx/repo-tools-data:repos.yaml


cpennington:
@edx/doc: Look ok to you folks?

openedx.yaml
```diff
@@ -0,0 +1,8 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: doc
+oeps: {}
+openedx-release: {ref: master}
+owner: edx/doc
+track-pulls: true
```
lamagnifica, line 8: no tags field, is this required per spec at http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification?

lamagnifica:
this looks good as far as I can tell, @cpennington . One question about whether a tags field is required, as inferred by the spec. One oddness in the spec is that it references both a openedx.yaml and openedx.yml file -- is there a reason for the difference? Do we actually need both in edx-documentation?

cpennington:
@lamagnifica: Nope, I need to fix the extension. Also, I've marked tags as optional in a pending PR: https://github.com/edx/open-edx-proposals/pull/18/files#diff-13c9b7b8133d4bcf6ab907c7f0065187R58.

PR #1195, opened 2016-08-22
Reduced fullstack installation instructions

nedbat:
There's no reason to have explicit instruction here, since they duplicate existing one.

### DIFF
--- a/en_us/install_operations/source/platform_releases/eucalyptus.rst
+++ b/en_us/install_operations/source/platform_releases/eucalyptus.rst
@@ -52,92 +52,14 @@ the Open edX Developer stack>` or :ref:`Fullstack <Installing Open edX Fullstack
 
 Review the prerequisites and instructions for each option, and then choose the
 option that best meets your needs. Ensure that you install the
-required software to run the edX platform.
+required software to run the Open edX platform.
 
 If you are upgrading from the Dogwood release, see `Upgrading from Dogwood to
 Eucalyptus`_.
 
-For new installations, follow these steps.
+Eucalyptus releases have Git tag names like ``open-release/eucalyptus.1``.
+The available names are detailed on the `Open edX Releases Wiki page`_.
 
-#. `Download the Vagrant Box`_ or `Download the BitTorrent File`_.
-
-   .. caution::
-     The Vagrant boxes have a large file size (between 4 and 5 GB). If you have
-     a slow or unreliable Internet connection, use BitTorrent to download the
-     Vagrant box you need.
-
-#. `Set the OPENEDX_RELEASE Environment Variable`_.
-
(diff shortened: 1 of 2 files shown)


nedbat:
@lamagnifica it feels a bit extreme, but I don't think we need those installation instructions there.  Not sure if I should add more though.

en_us/install_operations/source/platform_releases/eucalyptus.rst
```diff
@@ -52,92 +52,15 @@ the Open edX Developer stack>` or :ref:`Fullstack <Installing Open edX Fullstack
...
 option that best meets your needs. Ensure that you install the
-required software to run the edX platform.
+required software to run the Open edX platform.
 
 If you are upgrading from the Dogwood release, see `Upgrading from Dogwood to
 Eucalyptus`_.
 
-For new installations, follow these steps.
+Eucalyptus releases have Git tag names like ``open-release/eucalyptus.1``.
+The latest release is available as ``open-release/eucalyptus.latest``.  The
```
lamagnifica: for sure?
in your email earlier this evening, you include "The second release candidate is now available: open-release/eucalyptus.1rc2." Will "rc2" eventually fall off and leave "1"? will eucalyptus.latest = eucalyptus.1rc2 up until it turns into = eucalyptus.1?
I don't doubt you cn make this magic happen, just want to surface how I am interpreting these instructions.

lamagnifica:
I am a huge fan of providing instructions once, in one place. Enthusiastic 👍 . That said, I need some assurances about the way the release naming is presented. Thanks @nedbat !

en_us/install_operations/source/platform_releases/eucalyptus.rst
nedbat: Here is the tagging scheme we started using with Eucalyptus.  We should find a good place for it.

All of the Eucalyptus tags and branches start with `open-release/eucalyptus`, and a dot. Stable releases are numbered starting from 1: `open-release/eucalyptus.1`, `open-release/eucalyptus.2`, and so on.  Release candidates for those releases have "rc" suffixes, for example: `open-release/eucalyptus.1rc2`, `open-release/eucalyptus.3rc1`.  There are two branches which can refer to different code as time goes on: `open-release/eucalyptus.master` is the master branch where changes accumulate as we work toward the next release.  `open-release/eucalyptus.latest` is the latest stable release of Eucalyptus, or the latest release candidate if we haven't released `.1` yet.

Actually, as I type this, I'm less keen on the idea of `.latest`, since it allows people to not know what version they are installing, and may make it harder for us to record what they installed, and they'll need to look up specific details in some cases anyway... Hmmm.

PR #1206, opened 2016-08-26
Don't provide summary of upgrade process

nedbat:
Trying to explain the upgrade process is too involved, and will be
lopsided: there will be specifics, and then a step at the end that
amounts to, "and then all the rest of the upgrade" which is 95% of the
work.

So don't bother.


nedbat:
@lamagnifica looky

lamagnifica:
enthusiastic 👍

PR #1217, opened 2016-08-30
No need for $OPENEDX_RELEASE on install_stack.sh

nedbat:
I believe this is the only change on the Eucalyptus branch that isn't on master, that should be on master.


nedbat:
@lamagnifica OK?

lamagnifica:
@pdesjardins please take a look

pdesjardins:
From a comment in install_stack.sh: RELEASE is an optional positional argument, defaulting to OPENEDX_RELEASE.

It makes sense to me that the doc should not include that variable as a second argument, since it's the default.

👍

PR #1223, opened 2016-09-01
rm unnecessary 'cd devstack' step

cptvitamin:
the commands that follow are meant to be run above the devstack dir, not from within it. `bash install_stack.sh devstack`  installs devstack inside another devstack dir, so you end up with your devstack in devstack/devstack.  Furthermore, when you complete the script, you get a message that says `Finished installing! You may now 'cd devstack' and login using 'vagrant ssh'`  but you are already IN devstack.
### Reviewers

- [ ] @feanil 
- [ ] @nedbat

### DIFF
--- a/en_us/install_operations/source/installation/devstack/install_devstack.rst
+++ b/en_us/install_operations/source/installation/devstack/install_devstack.rst
@@ -110,7 +110,6 @@ To install devstack, follow these steps.
    .. code-block:: bash
 
      mkdir devstack
-     cd devstack
 
 #.  Set the ``OPENEDX_RELEASE`` environment variable to the Git tag name of the
     release of the Open edX platform that you are installing. For


feanil:
:+1:

en_us/install_operations/source/installation/devstack/install_devstack.rst
```diff
@@ -110,7 +110,6 @@ To install devstack, follow these steps.
    .. code-block:: bash
 
      mkdir devstack
-     cd devstack
```
nedbat, line 113: This won't work: the install script looks for an existing directory named devstack and refuses to run if it's there.

nedbat:
I think a better approach would be to change the install script so that it doesn't create the extra devstack directory.

nedbat:
This pull request changes the install script: https://github.com/edx/configuration/pull/3362

lamagnifica:
@nedbat so is this PR now something that should be closed unmerged? or should it go forward as it is? or is there some other change needed based on the PR you referenced?

nedbat:
The script now makes directories as @cptvitamin (and everyone else!) prefers, so no need for this change.

PR #1260, opened 2016-09-21
WIP: Reference `eucalyptus.2` tag instead of 1

gsong:
**WIP: Do no merge yet**

Refer to `eucalyptus.2` instead of `eucalyptus.1`.
### Reviewers

- [ ] Subject matter expert: @nedbat
### Testing
### Post-review
- [ ] Squash commits

### DIFF
--- a/en_us/install_operations/source/installation/devstack/install_devstack.rst
+++ b/en_us/install_operations/source/installation/devstack/install_devstack.rst
@@ -117,13 +117,13 @@ To install devstack, follow these steps.
     information about the latest Open edX releases and the Git tag names for
     them, see `Open edX Releases Wiki page`_.
 
-    For example, ``open-release/eucalyptus.1`` is the Git tag name for the
-    first Eucalyptus release. The following command sets the value of the
-    ``OPENEDX_RELEASE`` environment variable to ``open-release/eucalyptus.1``.
+    For example, ``open-release/eucalyptus.2`` is the Git tag name for the
+    second Eucalyptus release. The following command sets the value of the
+    ``OPENEDX_RELEASE`` environment variable to ``open-release/eucalyptus.2``.
 
     .. code-block:: bash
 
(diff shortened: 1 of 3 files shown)


lamagnifica:
@gsong that all looks fine to me. 👍  Something to consider for the future, though, is revising the way we present this information so that readers will go and find the most recent release version, rather than relying on us to update in in multiple places in multiple files?

nedbat:
We already have "For information about the latest Open edX releases and the Git tag names for them, see `Open edX Releases Wiki page`_."

I agree that we should not be updating this text for every release.  To that end, it might be best to make clear that the examples are fictitious. In other places, I've used "zebrawood.3" so that people won't be distracted by the "eucalyptus" word.

gsong:
@nedbat For those that just copy and paste, I wanted to make sure they get the latest release instead of an older release.

Shall we just close this PR then?

nedbat:
I intended "eucalyptus.1" to be an example, though it was a poor choice, since it was also the current release at the time.  What would make a clearer example?

catong:
@gsong What is the next step for this PR?

gsong:
@catong Haven't had a chance to come back to this. Can we leave it as WIP for now?

catong:
@gsong No problem! We are just checking on what might be inactive PRs. Thanks!

grantgoodmanedX:
We don't want to update the version with each release. Since it's just an example string, I like @nedbat 's tentative suggestion that we go with something like 'zebrawood.3'. I'd also be ok with just closing this PR, since there's nothing in the doc that's false, and the farther we get from eucalyptus.1, the less likely it is that a user will just cut and paste the example commands.

nedbat:
How about "examplewood.1" ? :)

gsong:
Since this hasn't been an issue, I'll just close out the PR.

PR #1262, opened 2016-09-21
DOC-3325: Remove references to the course structure API

lamagnifica:
## [DOC-3325](https://openedx.atlassian.net/browse/DOC-3325)

@nasthagiri requested a change from using the course structure API to the "course blocks API" to find usage IDs. Unfortunately, I was not able to use the courses API blocks resource for this purpose. Please provide minutely detailed instructions on how to accomplish the goal of using an API to find the usage IDs. An alternative is to remove this section on using an API and leave only the other two methods in the doc.
### Reviewers

- [x] Subject matter expert: @nasthagiri 
- [ ] Subject matter expert: 
- [ ] Doc team review (dev edit): @edx/doc
- [ ] Product review:
- [ ] Partner support: 
- [ ] PM review: 

FYI: @dsjen 
### Testing
### Post-review
- [ ] Squash commits

### DIFF
--- a/en_us/shared/course_features/lti/lti_address_content.rst
+++ b/en_us/shared/course_features/lti/lti_address_content.rst
@@ -94,27 +94,20 @@ indicate subsections in a course.
 Several methods are available to help you find the usage IDs for items in your
 course.
 
-To find the usage ID for a unit or a component in an edX course, you can use
-any of these methods.
+To find the usage ID for a unit or a component in the LMS, you can use
+either of these methods.
 
-* :ref:`View staff debug info<View Staff Debug Info for the Usage ID>` for it
-  in the edX LMS.
+* :ref:`View staff debug info<View Staff Debug Info for the Usage ID>` for the
+  unit or component.
 
-* :ref:`View the page source<View the Page Source for the Usage ID>`.
+* :ref:`View the page source<View the Page Source for the Usage ID>` for the
+  unit or component.
 
-* :ref:`View the course structure<View the Course Structure API for the Usage
-  ID>`.
+To find the usage ID for a subsection, you
+:ref:`view the page source<View the Page Source for the Usage ID>`.
 
-To find the usage ID for a subsection in an edX course, you can use one of
-these methods.
-
-* :ref:`View the page source<View the Page Source for the Usage ID>`.
-
-* :ref:`View the course structure API<View the Course Structure API for the
-  Usage ID>`.
-
-.. note:: You must have the Staff or Admin role in a course to follow any
- of these procedures for finding usage IDs.
+.. note:: You must have the Staff or Admin role in a course to follow these
+  procedures for finding usage IDs.
 
 .. _View Staff Debug Info for the Usage ID:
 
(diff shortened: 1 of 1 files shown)


en_us/shared/course_features/lti/lti_address_content.rst
```diff
@@ -205,40 +203,41 @@ and then copy and paste it into the spreadsheet.
...
+   For example, this URL accesses the edX Demo course.
 
-   ``https://edge.edx.org/api/course_structure/v0/course_structures/course-v1:edX+DemoX+Demo_Course``
+   ``https://edge.edx.org/api/courses/v1/blocks/?course_id=course-v1:edX+DemoX+Demo_Course``
 
-#. Press Enter. The course structure API appears in the browser.
+#. Press Enter. The courses API blocks resource appears in the browser.
 
-#. Scroll down to verify that an ``HTTP 200 OK`` message appears.
+.. Nimisha, I was unable to verify anything from this line down. Please advise. - Alison 21 Sept 16
```
nasthagiri: @lamagnifica Thanks for trying this out.  There are 2 changes that need to be made with the Course Blocks API.
1. The course-id needs to be URL encoded.  So instead of `course-v1:edX+DemoX+Demo_Course` it needs to be `course-v1%3AedX%2BDemoX%2BDemo_Course`.  One can use any URL encoding tool such as http://meyerweb.com/eric/tools/dencoder/ to URL encode a string.
2. The following 2 additional parameters need to be appended to the URL: `all_blocks=true` and `depth=all`

So, in your example, the new URL would be:
https://edge.edx.org/api/courses/v1/blocks/?course_id=course-v1%3AedX%2BDemoX%2BDemo_Course&all_blocks=true&depth=all

en_us/shared/course_features/lti/lti_address_content.rst
```diff
@@ -205,40 +203,41 @@ and then copy and paste it into the spreadsheet.
...
-   If you received a different HTTP response value, make sure that you have the
+#. Verify that an ``HTTP 200 OK`` message appears.
+
+   If you receive a different HTTP response value, make sure that you have the
    Staff or Admin role for the course, and that you have entered the URL
    correctly.
 
+.. I got a 400 response
+
 The API shows the ``root`` usage ID for your course, followed by the set of
```
nasthagiri: The rest remains the same, so no changes needed there.

en_us/shared/course_features/lti/lti_address_content.rst
lamagnifica: @nasthagiri sorry, I get further (I think), but I still get a 403 when logged in or a 401 when not logged in. How does one get credentials? Is it realistic to expect that a member of a course team could actually get a usage ID through this method? Should we just strip this section out, and leave the other two methods?

nasthagiri: @lamagnifica I think in this particular case, the other 2 methods are more than sufficient to get the usage key and it's not worth adding a 3rd way - especially if it leads to confusion.
However, I would like to understand why it's not working for you.  We can do so outside the scope of this PR.

nasthagiri:
👍

lamagnifica:
@nasthagiri thank you, let's touch base next week. New commit made.

lamagnifica:
doh! Quick Draw McGraw!

en_us/shared/course_features/lti/lti_address_content.rst
```diff
@@ -95,26 +95,17 @@ Several methods are available to help you find the usage IDs for items in your
 course.
 
 To find the usage ID for a unit or a component in an edX course, you can use
-any of these methods.
+either of these methods.
 
 * :ref:`View staff debug info<View Staff Debug Info for the Usage ID>` for it
```
catong: Suggest replace "it" here with "the unit or component"

en_us/shared/course_features/lti/lti_address_content.rst
```diff
@@ -95,26 +95,17 @@ Several methods are available to help you find the usage IDs for items in your
...
   in the edX LMS.
 
 * :ref:`View the page source<View the Page Source for the Usage ID>`.
```
catong: Optionally also add "for the unit or component" ("View the page source for the unit or component")

en_us/shared/course_features/lti/lti_address_content.rst
```diff
@@ -95,26 +95,17 @@ Several methods are available to help you find the usage IDs for items in your
...
 
-* :ref:`View the course structure<View the Course Structure API for the Usage
-  ID>`.
+To find the usage ID for a subsection in an edX course, you can
+:ref:`view the page source<View the Page Source for the Usage ID>`.
```
catong, line 107: Just my opinion, but "view the page source" needs "of the thing" appended, for clarity. So I'd recommend "you can view the page source of the subsection." or "you can view the subsection's page source."

catong (approved):
@lamagnifica Just a couple of suggestions around "view the page source (of the thing)", otherwise 👍. Progress!!

PR #1271, opened 2016-09-26
DOC-3253: post-CAPA a11y change cleanup

lamagnifica:
## [DOC-3253](https://openedx.atlassian.net/browse/DOC-3253)

As part of the work on changing markdown and OLX markup for a11y, Muzzafar's team produced a set of sample events. This PR does a careful check of their [samples](https://openedx.atlassian.net/wiki/display/AN/CAPA+events+after+a11y+changes) to the doc, and corrects deficiencies.
### Reviewers

- [x] Subject matter expert: @muhammad-ammar 
- [x] Subject matter expert: @stroilova 
- [ ] Doc team review (sanity check): @edx/doc
- [ ] Partner support: 
- [ ] PM review: 

FYI: @muzaffaryousaf 
### Testing
### HTML Version
- [ ] Here's an [HTML draft](http://draft-events-for-problems-20160926.readthedocs.io/en/latest/internal_data_formats/tracking_logs.html#problem-interaction-events) 
### Post-review
- [x] Squash commits

### DIFF
--- a/en_us/data/source/internal_data_formats/tracking_logs.rst
+++ b/en_us/data/source/internal_data_formats/tracking_logs.rst
@@ -2625,6 +2625,11 @@ hint.
 
 **History**: This event was added on 1 Jul 2015.
 
+``context`` **Member Fields**:
+
+This event type includes the :ref:`common<context>` ``context.module`` member
+field.
+
 ``event`` **Member Fields**:
 
 .. list-table::
@@ -2655,11 +2660,16 @@ hint.
 Course teams can design problems to include feedback messages that appear after
 a user submits an answer. For problems that include feedback messages, the
 server emits an ``edx.problem.hint.feedback_displayed`` event each time a user
-selects **Check**.
+selects **Submit**.
 
 **Event Source**: Server
 
 **History**: This event was added on 1 Jul 2015.
+
+``context`` **Member Fields**:
+
+This event type includes the :ref:`common<context>` ``context.module`` member
+field.
 
 ``event`` **Member Fields**:
 
@@ -2699,7 +2709,7 @@ selects **Check**.
        the specific problem for which the user received feedback.
    * - ``question_type``
      - string
-     - The XML tag that identifies the problem type. For example,
+     - The XML element that identifies the problem type. For example,
        'stringresponse' for a text input problem.
    * - ``student_answer``
      - array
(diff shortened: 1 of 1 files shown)


stroilova:
Thanks! Just one question about the "student" field -- what data element is this capturing (username, user_id, something else?)

muhammad-ammar:
@lamagnifica LGTM 👍 

BTW why we have changed `check` to `submit`? Isn't `check` more aligned with real word terminology?

lamagnifica:
@muhammad-ammar can you answer @stroilova 's question, "about the "student" field -- what data element is this capturing (username, user_id, something else?)" 
re: check vs. submit: Changes to the options that appear below CAPA problems in the LMS will be released next week. "Check" and "Final check" will be consolidated into a single button, relabeled "Submit".

muhammad-ammar:
@stroilova `student` key will be set to `username`

stroilova:
Thanks @muhammad-ammar ! Any reason why that field is called "student"? 
Should it be called username then, to be less ambiguous? 

If we do keep it as-is, we should update the docs to have an explanation of the field contents.

lamagnifica:
I'll add the explanation that the student field contains the username, thanks. 
Background:
This field is not new to the code, only to the doc. It's outside the scope of this effort to rename the field, or even to try to figure out why it was named the way it is. @stroilova if you feel strongly that the field should be renamed, just this one or possibly also where it occurs in other events, please open a JIRA story for an engineering team.

muhammad-ammar:
@stroilova Unfortunately i have no idea why this field is called `student`. It's added [automatically](https://github.com/edx/edx-platform/blob/master/lms/djangoapps/instructor_task/tasks_helper.py#L421) into the event info.

stroilova:
Thanks!  Usually we should try to use user_id, and have consistent naming. However, this was an older field, so let's keep as is.

Many thanks for the explanation @lamagnifica  and @muhammad-ammar 

👍

stroilova:
@muhammad-ammar  sorry, what do you mean "automatically"? Didn't someone on our side choose the field name at some point?

muhammad-ammar:
@stroilova `automatically` means that [student field](https://github.com/edx/edx-platform/blame/master/lms/djangoapps/instructor_task/tasks_helper.py#L421) is appended to actual event info from outside the CAPA. And yes you can change the field name. May be @brianhw can help understand why we are adding `student` field.

PR #1347, opened 2016-11-14
DOC-3495 Upgrade to Sphinx 1.3.5

jmbowman:
## [DOC-3495](https://openedx.atlassian.net/browse/DOC-3495)

Upgrade Sphinx to the current default version used by Read the Docs (which has been used for most recent builds anyway, since we didn't specify otherwise until today).  The version which had been in our requirements file was too old to support the ``-T`` parameter that Read the Docs passes to ``sphinx-build``.

### Reviewers

- [x] Doc team review (sanity check): @edx/doc
- [x] Product review: @nedbat 

### Testing

### Post-review

- [x] Squash commits

### DIFF
--- a/shared/travis_requirements.txt
+++ b/shared/travis_requirements.txt
@@ -1,5 +1,5 @@
 # Used for documentation gathering
 edx-sphinx-theme==1.0.2
-sphinx==1.1.3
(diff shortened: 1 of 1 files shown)


nedbat:
:+1:

PR #1352, opened 2016-11-22
DOC: Remove mention of practice proctored exam in DemoX

catong:
Removes mention of a practice proctored exam being in DemoX from the "online proctoring rules" topic that appears both in the _Learner's Guide_ and in the _Building and Running an edX Course_ guide.
FYI @jaakana


catong:
Merging https://github.com/edx/edx-documentation/pull/1352

PR #1372, opened 2016-12-14
Add note about selecting a release to ICR guide

srpearce:
## [DOC-3181](https://openedx.atlassian.net/browse/DOC-3181)

This PR adds a note about selecting a release to ICR guide.

### Reviewers

- [x] Doc team review (sanity check, copy edit, or dev edit?): @catong 

### Testing


catong:
@srpearce Just one fix, then 👍 In the file with the broken link, instead of having no link to the Progress page topic, could we replace the broken link with ```:ref:`check_student_progress` ```?

PR #1384, opened 2017-01-04
DOC-3527 Update Custom JS Example

catong:
## [DOC-3527](https://openedx.atlassian.net/browse/DOC-3527)

Updates the "Custom JavaScript Display and Grading Problem" with a new example problem and code. (TNL-5893, TNL-6121). Also includes changes made in TNL-6044 (add title attribute).

### Reviewers

- [x] Subject matter expert: @cahrens 
- [x] Doc team review: @srpearce  
- [x] Product review: @sstack22 

FYI: Partner support: @jaakana

### HTML Version for Review

- [x] http://draft-ca-customjsinput-example.readthedocs.io/en/latest/exercises_tools/custom_javascript.html?highlight=custom%20javascript#custom-javascript-display-and-grading-problem

### Testing

### Sandbox (from https://github.com/edx/edx-platform/pull/14072)
- [x] https://studio-jsinput.sandbox.edx.org/home/

### Post-review

- [x] Squash commits


en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -47,125 +45,133 @@ Create a Custom JavaScript Display and Grading Problem
...
+   is used for the title attribute on the generated inline frame.
+
+#. To add a **Save** button to your problem, select **Settings**, and then set
+   **Maximum Attempts** to a number larger than zero.
+
+#. Select **Save**.
+
 
-   All problems have more than one element. Most problems conform to the
+.. note::  All problems have more than one element. Most problems conform to the
```
cahrens: I'm not entirely sure what is meant by "all problems have more than one element". I _think_ this is referring to the fact that there are HTML elements outside of the iFrame that must be able to interact with content within the iFrame, but it is entirely clear to me.

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -47,125 +45,133 @@ Create a Custom JavaScript Display and Grading Problem
...
-#. If you want your problem to have a **Save** button, select **Settings**, and
-   then set **Maximum Attempts** to a number larger than zero.
-#. Select **Save**.
 
 ================================
 Recreate the Example Problem
 ================================
 
-To recreate the example problem above, you need the following files.
+To recreate the example problem above, you need the following files, which you can download from http://files.edx.org.
```
cahrens: These files are all within a directory. The actual URLs are

https://files.edx.org/custom-js-example/jsinput_example.html
https://files.edx.org/custom-js-example/jsinput_example.js
https://files.edx.org/custom-js-example/jsinput_example.css

This example also relies on this file--
https://files.edx.org/custom-js-example/jschannel.js

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -47,125 +45,133 @@ Create a Custom JavaScript Display and Grading Problem
...
-   - webGLDemo.js
-   - webGLDemo.css
-   - three.min.js
+.. note:: If you need to bypass the same-origin policy (SOP), you also need the
+   ``jschannel.js`` file from http://files.edx.org, and your
+   ``jsinput_example.html`` file will be slightly different.
 
-To download these files in a .zip archive, go to
-http://files.edx.org/JSInput.zip.
+#. Download the required files from http://files.edx.org.
```
cahrens: As long as you keep the path references in the example problem the same, there is no need to actually download these files and upload them. The example works as-is because the files live in a publicly accessible location.

The only reason to download the files would be to examine the contents as an example (which is useful when creating their own problems).

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -47,125 +45,133 @@ Create a Custom JavaScript Display and Grading Problem
...
+#. In Studio, upload the required files using the **Files & Uploads** page.
 
-.. note:: If you need to bypass the SOP, you also need the
-  ``jschannel.js`` file, and your webGLDemo.html file will be slightly
-  different. To download all of these files in a .zip archive, go to
-  http://files.edx.org/JSInput_BypassSOP.zip.
+#. In your course outline, create a new custom JavaScript display and grading
```
cahrens: The example problem will already have this code.

If you truly want to recreate this example "from scratch"--
[10:05 AM] Christina Roberts: If you are going to do that, it would make more sense to just create a blank advanced problem, then paste the code.
[10:05 AM] Christina Roberts: And you should change the example code in the documentation to use relative urls so that it uses the uploaded files instead of the one on files.edx.org.

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -47,125 +45,133 @@ Create a Custom JavaScript Display and Grading Problem
...
    you need to bypass the SOP. For example,
    ``https://info.company.com/JSInputElement2.html`` uses a different
-   protocol, host, and port. To bypass the SOP, change ``sop="false"`` in
-   line 8 of the example code to ``sop="true"``. For more information, see
-   the same-origin policy page on the `Mozilla Developer Network
-   <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Same_origin_policy_for_JavaScript>`_ or on
+   protocol, host, and port from
+   ``http://store.company.com:81/subdirectory_1/JSInputElement.html``.
+
+   To bypass the SOP, change ``sop="false"`` to ``sop="true"``. In the example
```
cahrens: This is actually the reverse of what it should be. sop="true" means that same-origin policy will be enforce, while false means that it is bypassed.

From the code-
// Bypass single-origin policy only if this attribute is "false"
// In that case, use JSChannel to do so.

Our example code disables sop and uses jschannel.

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -178,7 +184,7 @@ JSInput allows problem authors to turn stand-alone HTML files into problems
 that can be integrated into the edX platform. Since its aim is flexibility, it
 can be seen as the input and client-side equivalent of **CustomResponse**.
 
-A JSInput exercise creates an iframe in a static HTML page, and passes the
+A JSInput exercise creates an inline frame in a static HTML page, and passes the
```
cahrens: I would recommend putting iframe after "inline frame" in parenthesis.

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -214,6 +221,7 @@ set_statefn      Function name  No         None
 get_statefn      Function name  No         None
 height           Integer        No         `500`
 width            Integer        No         `400`
+title            String         No         None
```
cahrens, line 217: There actually is a default value of "Problem Remote Content". And although this is not technically required, for accessibility reasons it really should be set (so I would be tempted to just say it is required).

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -214,6 +221,7 @@ set_statefn      Function name  No         None
 get_statefn      Function name  No         None
 height           Integer        No         `500`
```
cahrens: Default in code is actually 300 (for height).

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -314,4 +322,8 @@ Optional Attributes
   file's dimensions (up to the aforementioned limits), but currently it
   defaults to `500` and `400` for **height** and **width**, respectively.
```
cahrens: 300 for height.

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -235,12 +243,12 @@ Required Attributes
...
-  refer to what `gradefn` expects.)
+  **myprog.myfun.call(myprog)**.
 
-  Aside from that, more or less anything goes. Note that currently there is no
-  support for inheriting CSS or JavaScript from the parent (aside from the
-  Chrome-only **seamless** attribute, which is set to True by default).
+  Other than the **gradefn** function, there are no other specific
+  requirements for the HTML file. Note that there is currently no support for
+  inheriting CSS or JavaScript from the parent (except for the Chrome-only
+  **seamless** attribute, which is set to ``True`` by default).
```
cahrens, line 251: Casing on "JavaScript" is incorrect here--

get_statefn

Sometimes the state and the answer are quite different. For instance, a problem that involves using a **javascript** program

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -33,9 +34,6 @@ Google Web Toolkit (GWT), to create your JS input problem.
     this option in the problem component, a **Show Answer** button appears in
     the learner's view of the problem in the LMS, but the button does not work.
 
-.. note::
```
cahrens, line 36: Thank you for deleting this!

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -47,125 +45,133 @@ Create a Custom JavaScript Display and Grading Problem
...
+   (**JSInputDemo.getGrade**, **JSInputDemo.getState**, and
+   **JSInputDemo.setState**).
+
+ - The JavaScript input problem code uses **JSInputDemo.getGrade**,
+   **JSInputDemo.getState**, and **JSInputDemo.setState** to grade, save, or
+   restore a problem. These functions must be global in scope.
 
- - **Width** and **height** represent the dimensions of the iframe that holds
-   the application.
+ - **JSInputDemo.getGrade** and **JSInputDemo.setState** are optional. You
```
cahrens: Should be getState and setState, not getGrade.

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -235,12 +243,12 @@ Required Attributes
   uses the **gradefn** function, `gradefn` is called with
   `gradefn`.call(`obj`), where **obj** is the object-part of **gradefn**. For
   example, if **gradefn** is **myprog.myfn**, JSInput calls
-  **myprog.myfun.call(myprog)**. (This is to ensure "`this`" continues to
-  refer to what `gradefn` expects.)
+  **myprog.myfun.call(myprog)**.
 
-  Aside from that, more or less anything goes. Note that currently there is no
```
cahrens, line 241: You don't like "anything goes"? :)

cahrens:
@catong I'm done with this pass.

en_us/shared/exercises_tools/custom_javascript.rst
catong: @cahrens does it make sense to you that what I think of as URLs are called "elements" in the following sentences? Seems that the point of the paragraph is to explain the SOP, but perhaps the term used ("element") is confusing. What would you call them?

en_us/shared/exercises_tools/custom_javascript.rst
catong: Is it true that the jschannel.js file is only required if you want to bypass the SOP? (line 93-95 below)

cahrens: Yes, I think that is true.

en_us/shared/exercises_tools/custom_javascript.rst
cahrens: Honestly, I still don't understand what "elements" is referring to. I wouldn't say that a jsinput problem is made up of multiple URLs. 

Is it referring to origins?

catong: The original text says "all problems *have more than one* (rather than "are made up of multiple"). I don't understand what you mean by "origins", but if you were to say "A Custom JSInput problem will have more than one ```<blah>```, and if those ```<blah>```s do not have the same protocol, host, and port, then you need to define ```sop=false```", what term would be appropriate for "blah"?

cahrens: How about you change "element" to "resource" throughout the section. I got resource from the language in https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy.

catong: That works! Thanks!

sstack22:
LGTM, I defer to @cahrens for final thumbs.

en_us/shared/exercises_tools/custom_javascript.rst
catong: @cahrens Where does this height default live? I can't find height = 300 in the yaml file or anywhere in your PR, or in the live template on Prod.

catong: @cahrens Where does the default height live? I couldn't find ```height = 300``` in the yaml file, in the used files, in your PR, or in the live template on Prod.

cahrens: It's in the Python code, if no height is set on the jsinput HTML element.

````
    @classmethod
    def get_attributes(cls):
        """
        Register the attributes.
        """
        return [
            Attribute('params', None),       # extra iframe params
            Attribute('html_file', None),
            Attribute('gradefn', "gradefn"),
            Attribute('get_statefn', None),  # Function to call in iframe
                                             #   to get current state.
            Attribute('initial_state', None),  # JSON string to be used as initial state
            Attribute('set_statefn', None),  # Function to call iframe to
                                             #   set state
            Attribute('width', "400"),       # iframe width
            Attribute('height', "300"),      # iframe height
            # Title for the iframe, which should be supplied by the author of the problem. Not translated
            # because we are in a class method and therefore do not have access to capa_system.i18n.
            # Note that the default "display name" for the problem is also not translated.
            Attribute('title', "Problem Remote Content"),
            # SOP will be relaxed only if this attribute is set to false.
            Attribute('sop', None)
        ]

File: en_us/shared/exercises_tools/custom_javascript.rst [hunk 14]
```diff
@@ -56,59 +56,46 @@ Create a Custom JavaScript Display and Grading Problem
...
    protocol, host, and port from
    ``http://store.company.com:81/subdirectory_1/JSInputElement.html``.
 
-   To bypass the SOP, change ``sop="false"`` to ``sop="true"``. In the example
+   To bypass the SOP, change ``sop="true"`` to ``sop="false"``. In the example
    problem code, this attribute is just before the closing ``customresponse``
    tag.
 
+   If you bypass the same-origin policy (SOP), you require an additional file.
+   The example problem uses the file ``JSChannel.js`` to bypass the SOP.
```
cahrens: Should be lower-case (jschannel.js).

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -202,11 +190,11 @@ The following is the basic format of a JSInput problem.
         </script>
         <customresponse cfn="all_true">
             <jsinput gradefn="gradefn"
-                height="500"
+                height="300"
```
cahrens: There's no need to change this to 300 (in fact, it doesn't make much sense because that's the default in Python code if no height is set). I would leave it as 500.

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -202,11 +190,11 @@ The following is the basic format of a JSInput problem.
...
                 get_statefn="getstate"
                 set_statefn="setstate"
                 html_file="/static/jsinput.html"
-                title="iframe Title"/>
+                title="Problem Remote Content"/>
```
cahrens: Same comment as above-- no reason to duplicate the default value from Python code here.

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -47,125 +45,120 @@ Create a Custom JavaScript Display and Grading Problem
...
+   return answer == "correct"
+
+   # Or you can use the value of the state key to grade:
+   """
+   state = json.loads(response["state"])
+   return state["selectedChoice"] == "correct"
+   """
+ ]]>
+ </script>
+     <p>This is paragraph text displayed before the iframe.</p>
```
cahrens: If there is a way to fix the indenting of this file, that would be great. The closing tag of "customresponse" should be at the same indent level as the opening tag (line 102). Here is what it should look like (with an indent level 4, you could also do an indent level of 2 if you think this takes up too much room)--


````
<problem>
    <customresponse cfn="check_function">
        <script type="loncapa/python">
            <![CDATA[
            import json
            def check_function(e, ans):
                """
                "response" is a dictionary that contains two keys, "answer" and "state".
                The value of "answer" is the JSON string that "getGrade" returns.
                The value of "state" is the JSON string that "getState" returns.
                Clicking either "Submit" or "Save" registers the current state.
                """
                response = json.loads(ans)

                # You can use the value of the answer key to grade:
                answer = json.loads(response["answer"])
                return answer == "correct"

                # Or you can use the value of the state key to grade:
                """
                state = json.loads(response["state"])
                return state["selectedChoice"] == "correct"
                """
            ]]>
        </script>
        <p>This is paragraph text displayed before the iframe.</p>
        <jsinput
            gradefn="JSInputDemo.getGrade"
            get_statefn="JSInputDemo.getState"
            set_statefn="JSInputDemo.setState"
            initial_state='{"selectedChoice": "incorrect1", "availableChoices":  ["incorrect1", "correct", "incorrect2"]}'
            width="600"
            height="100"
            html_file="https://files.edx.org/custom-js-example/jsinput_example.html"
            title="Dropdown with Dynamic Text"
            sop="false"/>
    </customresponse>
</problem>

**Comment by catong** [comment]
@cahrens I made your 2nd round changes; ready for your re-review. Thanks!

**Comment by catong** [comment]
@srpearce Also ready for your review. Thanks!

File: en_us/shared/exercises_tools/custom_javascript.rst [hunk 18]
```diff
@@ -99,46 +99,50 @@ overview. The example problem uses these files.
...
-        html_file="https://files.edx.org/custom-js-example/jsinput_example.html"
-        title="Dropdown with Dynamic Text"
-        sop="false"/>
+     <customresponse cfn="check_function">
+         <script type="loncapa/python">
+             <![CDATA[
+             import json
+             def check_function(e, ans):
+                 """
+
```
cahrens: Extra line?

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -199,7 +197,8 @@ The following is the basic format of a JSInput problem.
                 height="500"
```
cahrens, line 197: Nit: can you fix the indenting on line 193?

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -209,11 +208,13 @@ The accepted attributes are:
 Attribute Name   Value Type     Required   Default
 ==============  ==============  =========  ==========
 html_file        URL string     Yes        None
+title            string         Yes        Problem Remote Content
```
cahrens: Should this be in italics? Default values for gradefun, height, and width are (though I'm not sure why).

cahrens:
👍 Just a few more minor things, but otherwise looks great!

en_us/shared/exercises_tools/custom_javascript.rst
catong: As actual values they are meant to appear in code/monospace font. I fixed the formatting so that all the default values including the title default appear in monospace.

catong:
@cahrens Thanks for the careful review. I addressed your final comments.

cahrens:
Thanks @catong -- 👍

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -19,7 +19,8 @@ style sheets (CSS). You can use any application creation tool, such as the
 Google Web Toolkit (GWT), to create your JS input problem.
 
 .. image:: ../../../shared/images/JavaScriptInputExample.png
- :alt: Image of a JavaScript Input problem
+ :alt: An example JavaScript Input problem that contains a dropdown selector
+       inside an inline frame
```
srpearce: Add period

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -47,125 +45,123 @@ Create a Custom JavaScript Display and Grading Problem
    Component** select **Problem** , and then select **Advanced**.
 #. Select **Custom JavaScript Display and Grading**.
 #. In the component that appears, select **Edit**.
+
 #. In the component editor, modify the example code according to your problem.
+   Be sure to specify a ``title`` attribute on the ``jsinput`` tag. This title
+   is used for the title attribute on the generated inline frame.
+
+#. To add a **Save** button to your problem, select **Settings**, and then set
```
srpearce: If this step is optional, I'd add (optional) after the step number.

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -47,125 +45,123 @@ Create a Custom JavaScript Display and Grading Problem
...
+   ``http://store.company.com:81/subdirectory_1/JSInputElement.html``.
+
+   To bypass the SOP, change ``sop="true"`` to ``sop="false"``. In the example
+   problem code, this attribute is just before the closing ``customresponse``
+   tag.
+
+   If you bypass the same-origin policy (SOP), you require an additional file.
+   The example problem uses the file ``jschannel.js`` to bypass the SOP.
+
+   For more information, see the same-origin policy page on the `Mozilla
```
srpearce, line 79: _in_ the Mozilla Developer Network? "On" looks a bit odd to me.

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -47,125 +45,123 @@ Create a Custom JavaScript Display and Grading Problem
...
-   the same-origin policy page on the `Mozilla Developer Network
-   <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Same_origin_policy_for_JavaScript>`_ or on
+   protocol, host, and port from
+   ``http://store.company.com:81/subdirectory_1/JSInputElement.html``.
+
+   To bypass the SOP, change ``sop="true"`` to ``sop="false"``. In the example
+   problem code, this attribute is just before the closing ``customresponse``
+   tag.
+
+   If you bypass the same-origin policy (SOP), you require an additional file.
```
srpearce: No need to redefine the acronym here.

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -47,125 +45,123 @@ Create a Custom JavaScript Display and Grading Problem
...
+========================================
+
+The following code recreates the JavaScript Input problem example shown in the
+overview. The example problem uses these files.
+
+* https://files.edx.org/custom-js-example/jsinput_example.html
+* https://files.edx.org/custom-js-example/jsinput_example.js
+* https://files.edx.org/custom-js-example/jsinput_example.css
+* https://files.edx.org/custom-js-example/jschannel.js (This file is used only
+  because ``sop=false``)
```
srpearce: This seems like a shorthand way to say this. Maybe:

This file is used only because this example bypasses the SOP, as indicated by the `sop="false"` line.

(Note: "false" in the code should be surrounded with quotation marks)

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -178,9 +174,10 @@ JSInput allows problem authors to turn stand-alone HTML files into problems
 that can be integrated into the edX platform. Since its aim is flexibility, it
 can be seen as the input and client-side equivalent of **CustomResponse**.
 
-A JSInput exercise creates an iframe in a static HTML page, and passes the
-return value of author-specified functions to the enclosing response type
-(generally **CustomResponse**). JSInput can also store and retrieve state.
+A JSInput exercise creates an inline frame (iframe) in a static HTML page, and
+passes the return value of author-specified functions to the enclosing
+response type (generally **CustomResponse**). JSInput can also store and
+retrieve state.
```
srpearce, line 181: This seems like a bit late to include this information. Should it be in the problem description at the beginning of the topic?

en_us/shared/exercises_tools/custom_javascript.rst
```diff
@@ -235,12 +235,18 @@ Required Attributes
...
   `gradefn`.call(`obj`), where **obj** is the object-part of **gradefn**. For
   example, if **gradefn** is **myprog.myfn**, JSInput calls
-  **myprog.myfun.call(myprog)**. (This is to ensure "`this`" continues to
-  refer to what `gradefn` expects.)
+  **myprog.myfun.call(myprog)**.
+
+  Other than the **gradefn** function, there are no other specific
+  requirements for the HTML file. Note that there is currently no support for
+  inheriting CSS or JavaScript from the parent (except for the Chrome-only
+  **seamless** attribute, which is set to ``True`` by default).
```
srpearce: Nit: I'd recast both "there are" constructions—especially the second one, since the entity who could be offering the support is unclear. 

... the HTML file has no specific requirements. Note that edX does not currently support inheriting CSS...

srpearce (commented):
A few comments. Looking good!

catong:
Merging https://github.com/edx/edx-documentation/pull/1384

PR #1386, opened 2017-01-06
Insights and Data API release notes for 2017-01-09

thallada:
We released a new version of Insights and the Analytics Data API today to prod and edge and would like to include the changes in the next release notes update.

### Date Needed (optional)

We released it today, so by the next release notes update would be best.

### Reviewers

- [x] Subject matter expert: @dsjen 
- [x] Doc team review (sanity check and copy edit): @edx/doc
- [ ] Product review: @katymyw

FYI: Tag anyone else who might be interested in this PR here. @stroilova 

### Testing

### HTML Version (optional)

### Sandbox (optional)

### Post-review

- [ ] Squash commits


en_us/release_notes/source/2017/analytics/analytics_2017-01-09.rst
```diff
@@ -0,0 +1,9 @@
+Update "student" to "learner" in the Insights guide text.
+
+Fixed a bug that caused Insights to throw a 500 error when a site-wide
```
dsjen: Nit: I think that this note might not be needed; it's sort of user-facing, but not totally relevant?

dsjen (approved):
Just one question about whether to include a point (no strong preferences).  LGTM!

en_us/release_notes/source/2017/analytics/analytics_2017-01-09.rst
thallada: True, it's only relevant to Open Edx developers that may have encountered the bug. I'll omit this one.

catong:
Doc 👍

katymyw:
small question -- do we want to say "attempt to fix" in the release notes? (I know we're not certain). Probably fine to leave as is, but just wanted to ask. 

otherwise, 👍

thallada:
@katymyw Yeah I wasn't totally sure what to put for that. We never confirmed that it actually fixed the bug the user was seeing, so it is an "attempt" still. Perhaps could find a better wording for that...

PR #1398, opened 2017-01-18
DOC-3546 Support track-based bulk email groups

catong:
## [DOC-3546](https://openedx.atlassian.net/browse/DOC-3546)

Updates the topic about bulk email in the "Building and Running an edX Course" guide to include mention of track-based recipient groups. (Added in TNL-6162, https://github.com/edx/edx-platform/pull/14118)

### Date Needed: Jan 19
Should be live in Production on Jan 19

### Reviewers

- [x] Subject matter expert: @efischer19 
- [x] Doc team review (sanity check, copy edit, or dev edit?): @srpearce 
- [x] Product review: @sstack22 @marcotuts 

FYI: @jaakana, @mmacfarlane, @dhixonedx

### Testing

### Post-review

- [x] Add a comment with the description of this change or link this PR to the next release notes task.
- [x] Squash commits

### DIFF
--- a/en_us/shared/manage_live_course/bulk_email.rst
+++ b/en_us/shared/manage_live_course/bulk_email.rst
@@ -61,9 +61,12 @@ The following preset recipient groups are available.
   does not include learners who have not activated their accounts, or who
   have opted out of receiving email communications.
 
-In addition, if you have cohorts enabled in your course, each cohort is
-available as a separate recipient group. For more information, see :ref:`Bulk
-Email Cohorts`.
+If you have more than one enrollment track in your course, each enrollment
+track is available as a separate recipient group. For more information, see
+:ref:`Enrollment Track Recipient Groups`.
+
+If you have cohorts enabled in your course, each cohort is available as a
+separate recipient group. For more information, see :ref:`Bulk Email Cohorts`.
 
 
 .. _Bulk Email Who Is Included or Excluded:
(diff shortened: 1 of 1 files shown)


en_us/shared/manage_live_course/bulk_email.rst
```diff
@@ -106,6 +109,22 @@ should be aware of who is included in each group.
...
 
+.. _Enrollment Track Recipient Groups:
+
+==================================================================
+Sending Email Messages to Learners in Different Enrollment Tracks
+==================================================================
+
+If you have more than one enrollment track in your course, each enrollment
+track is available as a separate recipient group. If your course includes only
+a single default enrollment track, you will not have a track-based recipient
```
catong: @efischer19 Is this accurate re "default" enrollment track? Could I say "only a single enrollment track"? or is there always a default one?

efischer19: I'm not super familiar with how this has been described in the past, but `only a single enrollment track` sounds accurate to me.

efischer19 (approved):

en_us/shared/manage_live_course/bulk_email.rst
catong: Great, thanks! I will make the edit.

marcotuts:
👍

srpearce (approved):

JAAkana:
The doc changes are clear!

Q. about the functionality: I see that you can contact learners by enrollment track, and by cohort. Will courses with automated Verified Cohorting see this group of learners listed only once in the list of Recipients, or twice redundantly (e.g. listed once as Learners in the Verified Certificate Track, and once as "name of verified cohort")

catong:
@JAAkana I *think* that both the cohort and the enrollment track will each be available as an addressing group option, because each of these groups exists, and the logic doesn't know that in a verified track cohort implementation, the set of learners in both groups should be identical.  
@efischer19 is this correct?

efischer19:
@catong @JAAkana that is correct. The bulk email options are "dumb", in that they aren't aware of the verified track cohorting feature. So the cohort and the track will be listed as options, but if both are selected each learner will still only receive one email due to our de-duplicating logic on the backend (which also catches the case where Student A is in cohort B and registration track C, and the staff user selected both B and C as targets).

catong:
Thanks for confirming, @efischer19. @JAAkana are you fine with the doc as is? When creating this PR I did consider mentioning more use cases, but ultimately decided that the existing comment about de-duplication was sufficient, and that more explanation might actually be more confusing. Let me know if you think otherwise.

JAAkana:
Perfect, thank you @efischer19 and @catong  - Thumbs up on the doc as-is, I agree that my case is a bit of an edge case at the moment and we should avoid potential confusion.

catong:
Merging https://github.com/edx/edx-documentation/pull/1398

PR #1404, opened 2017-02-01
Fix cross references to use renamed anchor

catong:
Fix cross references to use new name for existing anchor
FYi @srpearce


catong:
Merging https://github.com/edx/edx-documentation/pull/1404

PR #1406, opened 2017-02-03
Add new anchors to proctoring topic

catong:
Add anchors to existing sections in SFD proctoring results topic, to facilitate linking from proctoring results email that gets sent to learners.
 FYI @edx/doc


catong:
Merging https://github.com/edx/edx-documentation/pull/1406

PR #1422, opened 2017-02-22
Change inline code indicator to prevent build errors

srpearce:
This PR changes the ``.. code-block:: python`` directives to just ``::`` to prevent the following build error:

``WARNING: Could not lex literal_block as "python". Highlighting skipped.``


catong:
👍

PR #1443, opened 2017-03-13
Describe assignment type average as weighted

grantgoodmanedX:
## [DOC-3596](https://openedx.atlassian.net/browse/DOC-3596)

The instructor's grade report includes average grades by assignment type. This average is now weighted by the assignment type weight, so I mentioned that at the two points we discuss the average.

### Reviewers

- [ ] Subject matter expert: @sstack22 : 
- [x] Doc team review (copy edit): @edx/doc

### Testing

### Post-review

- [ ] Squash commits

### DIFF
--- a/en_us/shared/student_progress/course_grades.rst
+++ b/en_us/shared/student_progress/course_grades.rst
@@ -184,11 +184,13 @@ columns that provide the following information.
   that assignment type: for example, "Homework (Avg)". This column is not
   included if a particular assignment type has only one assignment.
 
-  Note that this assignment type average takes into account dropped
-  assignments. For example, if the course includes five homework assignments
-  and the course grading policy allows one homework assignment with the lowest
-  score to be dropped, the homework assignment average in this grade report is
-  calculated over four homework assignments rather than five.
+  This assignment type average takes both dropped assignments and the assignment
+  weight into account. For example, if the course includes five homework
+  assignments and the course grading policy allows one homework assignment with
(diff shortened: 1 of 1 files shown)


en_us/shared/student_progress/course_grades.rst
```diff
@@ -184,11 +184,14 @@ columns that provide the following information.
...
-  score to be dropped, the homework assignment average in this grade report is
-  calculated over four homework assignments rather than five.
+  This assignment type average is modified by the assignment weight. In other
+  words, the assignment type average is the average of all assignment scores of
+  that type, multiplied by the weight of that assignment type. The average also
+  takes into account dropped assignments. For example, if the course includes
+  five homework assignments and the course grading policy allows one homework
+  assignment with the lowest score to be dropped, the homework assignment
+  average in this grade report is calculated over four homework assignments
+  rather than five.
```
srpearce: I would recast this a bit to emphasize the weighted aspect more, since that seems to be what was confusing course teams. Maybe:

This assignment type average takes both dropped assignments and the assignment weight into account. For example, if the course includes five homework assignments and the course grading policy allows one homework assignment with the lowest score to be dropped, the homework assignment average in this grade report is calculated over four homework assignments rather than five. This average is then multiplied by the assignment weight to calculate the assignment type average.

grantgoodmanedX: That's a distinct improvement, thanks.

PR #1456, opened 2017-03-29
SEVERE errors should count toward errors

nedbat:
## Count SEVERE errors also

Some errors are marked "SEVERE", like headings with the wrong markers. Those errors weren't counted by run_tests.sh.  This issue is now fixed.

### Reviewers

- [ ] Subject matter expert: @jmbowman 
- [ ] Doc team review (just cc): @edx/doc

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -72,11 +72,8 @@ for project in "${projects[@]}"; do
         BUILD_ERRORS=$((BUILD_ERRORS + 1))
     fi
 
-    # Get errors and warnings from the error log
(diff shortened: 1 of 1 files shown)


nedbat:
@jmbowman ping

jmbowman (approved):
👍  Definitely simpler this way too.  Do you know if there was a particular reason for using the original read + grep setup?

Very weird that docutils uses "severe" while logging uses "critical"; the respective PEPs even came out within a year of each other (258 and 282).

nedbat:
I can only assume the convoluted original code was due to sequential google-cut-paste cycles :)

PR #1498, opened 2017-05-05
Move LMS instructor dashboard Help topic into B&R guides

catong:
## [DOC-3411](https://openedx.atlassian.net/browse/DOC-3411)

This PR creates a landing page within the Building and Running guide for the Help link in the LMS instructor dashboard. 
(Linking to individual topics is not possible until instructor dashboard tabs have individual help context - for now, there's just one Help context regardless of which tab you are on, hence this landing page.)
This PR is the first of 2 steps to move the Help link - first we need to publish this new page, then we will update the Help config in edx-platform to point to this topic.

### Reviewers

Possible roles follow. The PR submitter checks the boxes after each reviewer finishes and gives :+1:. 
- [x] Doc team review (sanity check): @edx/doc 

### Testing

- [x] Ran ./run_tests.sh - known error caused by new anchors

### Post-review

- [ ] Add a comment with the description of this change or link this PR to the next release notes task.
- [x] Squash commits

PR #1499, opened 2017-05-09
DOC-3411 - Add missing anchor

catong:
Add anchor missing from https://github.com/edx/edx-documentation/pull/1498


catong:
Merging https://github.com/edx/edx-documentation/pull/1499
Known issue: test will fail until guide is republished with new anchor

PR #1530, opened 2017-06-28
Fix typo that caused note to display incorrectly in Publisher topics

srpearce:
Fixes a missing space that causes the note at the top of each page in the Publisher topic to display incorrectly.

### Date Needed (optional)
ASAP

- [ ] Doc team review (sanity check): @edx/doc


catong:
👍

PR #1549, opened 2017-07-24
Fix XBlock SDK database migration

regisb:
The latest XBlock SDK version (v0.1.4) is compatible with Django 1.11,
but not Django 1.8. Consequently, the `syncdb` command will not work
with v0.1.4. To address this problem in the Open edX documentation, we
choose to stick with the same Django version as Ficus/Ginkgo.

Affected user: https://openedx.slack.com/archives/C0GF6FTHA/p1500890543190218
(I know another faced this problem before, but I can't remember who)

### Date Needed

It would be great if this change was pushed to the official documentation as soon as possible, because XBlock developer currently can't create new XBlocks by following the documentation. The error that they obtain is cryptic ("no such table: workbench_xblockstate").

### Reviewers

- [ ] Subject matter expert: @mhoeber (wrote the rest of the documentation page)

### Testing

```
********** All builds done ************
TOTAL SPHINX ERRORS: 0
TOTAL SPHINX WARNINGS: 8
OTHER BUILD ERRORS: 0
There were problems while building the following projects:
en_us/course_authors
en_us/open_edx_course_authors
```

### Post-review

- [ ] Squash commits

### DIFF
--- a/en_us/xblock-tutorial/source/reusable/create_db.rst
+++ b/en_us/xblock-tutorial/source/reusable/create_db.rst
@@ -6,11 +6,11 @@ Before running the XBlock SDK the first time, you must create the SQLite
 database.
 
 #. In the ``xblock_development`` directory, run the following command to create
-   the database.
+   the database and the tables.
 
    .. code-block:: none
 
-      (venv) $ python xblock-sdk/manage.py syncdb
+      (venv) $ python xblock-sdk/manage.py migrate
 
 #. You are prompted to indicate whether or not to create a Django superuser.
 


openedx-webhooks:
Thanks for the pull request, @regisb! I've created [OSPR-1843](https://openedx.atlassian.net/browse/OSPR-1843) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.

efischer19 (changes requested):
Per [django docs](https://docs.djangoproject.com/en/1.7/ref/django-admin/#syncdb), `syncdb` has been a simple alias to `migrate` since 1.7. I think a better fix for this issue would be to `s/syncdb/migrate` in  https://github.com/edx/edx-documentation/blob/master/en_us/xblock-tutorial/source/reusable/create_db.rst.

This will allow developers to continue using the latest version, which should work correctly with both django 1.8 and 1.11.

regisb:
@efischer19 `syncdb` is no longer an alias for `migrate` in Django 1.11:
https://docs.djangoproject.com/en/1.11/ref/django-admin/#syncdb

The command is no longer available: https://github.com/django/django/tree/master/django/core/management/commands

I have considered modifying the documentation to rely on django 1.11 instead of 1.8, but that would mean that developers would create XBlocks with the django 1.11 API, whereas edx-platform still relies on django 1.8. This could cause XBlocks to work in the workbench, but not with Ficus/Ginkgo.

efischer19:
Right, it *was* an alias, it is no longer. At any point during 1.8's usage, we can update the documentation to say `migrate` instead of `syncdb`; this will not change anything. As of 1.11, `syncdb` is no longer supported, so we *must* make the harmless change.

xblock-sdk, along with a lot of other repos, is designed to work with a range of django versions. The documentation should rely on django, but not on a specific version being installed. Thus, we should no longer refer to the 1.8 exclusive `syncdb` command and use the identical, but available in the full range of supported versions, `migrate`.

You're right in that a developer *could* make an XBlock that relies on some new, "only-available-in-1.11" feature using the latest version of xblock-sdk, and that XBlock would not work with Ficus/Ginkgo. However, the vast majority will work correctly in both versions, and updating the range of allowed versions in xblock-sdk now makes the work to update edx-platform to 1.11 much easier in the future.

regisb:
You're right @efischer19, I missed the point in my previous message. Since `migrate` is common to Django 1.8 and 1.11, it makes more sense to use `migrate` than `syncdb`.
I'll test the install procedure with v0.1.4 and update this PR.

regisb:
Any chance to see this merged any time soon? Unless I'm mistaken, the travis build warnings are not related to this commit.

efischer19 (approved):
Sorry for losing track of this one @regisb! I agree that the failures are unrelated and we ought to merge this, but I'll ask someone @edx/doc to verify that.

The warnings look like this, but `en_us/xblock-tutorial/source/reusable/create_db.rst` is the only file touched by this PR.

```
../shared/course_features/cohorts/cohorted_courseware.rst:144: WARNING: undefined label: specify components in courseware as available only to certain content groups (if the link has no caption the label must precede a section header)
../shared/course_features/cohorts/cohorted_courseware.rst:223: WARNING: undefined label: specify components in courseware as available only to certain content groups (if the link has no caption the label must precede a section header)
../shared/developing_course/workflow.rst:94: WARNING: undefined label: content groups (if the link has no caption the label must precede a section header)
../shared/developing_course/workflow.rst:94: WARNING: undefined label: enrollment track groups (if the link has no caption the label must precede a section header)
```

catong:
@regisb Can you please rebase against master? The errors should clear after you do that.

regisb:
Awesome, thanks!

catong:
@regisb Thank you very much for your contribution!

PR #1558, opened 2017-08-04
Grant/ginkgo

grantgoodmanedX:
Ginkgo versions of Open edX docs

### DIFF
--- a/en_us/install_operations/source/platform_releases/ginkgo.rst
+++ b/en_us/install_operations/source/platform_releases/ginkgo.rst
@@ -1,0 +1,200 @@
+.. _Open edX Ginkgo Release:
+
+########################
+Open edX Ginkgo Release
+########################
+
+This section describes how to install the Open edX Ginkgo release.
+
+.. contents::
+ :local:
+ :depth: 1
+
+**************************
+What's Included in Ginkgo
+**************************
+
+The Open edX Ginkgo release contains several new features for learners,
+course teams, and developers. For more information, see
+:ref:`openreleasenotes:Open edX Ginkgo Release`.
+
+****************************
+What Is the Ginkgo Git Tag?
+****************************
+
+A git tag identifies the version of Open edX code that is the Ginkgo release.
+You can find the most up-to-date git tag for the current Open edX release on
+the `Open edX Releases Wiki page`_.
+
+The following Open edX git repositories have the Ginkgo git tag:
+
+* edx-platform
+* configuration
+* course_discovery
+* cs_comments_service
+* xqueue
+* ecommerce
+* ecommerce-worker
+* edx-analytics-configuration
+* edx-analytics-dashboard
+* edx-analytics-data-api
+* edx-analytics-pipeline
+* edx-certificates
+* edx-custom-a11y-rules
+* edx-demo-course
+* edx-documentation
+* edx-notes-api
+* edx-ui-toolkit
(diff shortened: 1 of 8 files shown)


en_us/install_operations/source/platform_releases/ginkgo.rst
```diff
@@ -0,0 +1,103 @@
...
+* edx-analytics-data-api
+* edx-analytics-pipeline
+* edx-certificates
+* edx-custom-a11y-rules
+* edx-demo-course
+* edx-documentation
+* edx-notes-api
+* edx-ui-toolkit
+* notifier
+* programs
```
nedbat: Remove programs.

en_us/install_operations/source/platform_releases/ginkgo.rst
```diff
@@ -0,0 +1,103 @@
...
+****************************
+
+A git tag identifies the version of Open edX code that is the Ginkgo release.
+You can find the most up-to-date git tag for the current Open edX release on
+the `Open edX Releases Wiki page`_.
+
+The following Open edX git repositories have the Ginkgo git tag:
+
+* edx-platform
+* configuration
```
nedbat, line 32: Add "course_discovery"

nedbat (dismissed):

grantgoodmanedX:
@nedbat I added a migration section in this PR, rather than the release note PR, since they are on different branches. I have a [RUN SOME SCRIPT] placeholder that might be where you add your migration script and instructions. I'll also need information about how the component migration steps fit in with an overall migration script, if at all.

nedbat:
@feanil @jibsheet Would you mind taking a quick look at the upgrade instructions? I did a bunch of bash to make dumping the database smooth.

jibsheet:
@nedbat This looks good to me, @feanil has a few questions that cover both of us.

feanil (approved):

feanil:
@nedbat I had some questions around mongo indexes. If we add or update indexes between two releases, we would want to point them to docs for the changed indexes. I don't think that happened here but just making sure.  Also, by default our code doesn't create indexes automatically on mongo, do we need to add docs for how to create the default indexes here as well?

nedbat:
@feanil @jibsheet I'm looking to you for expertise on this.  If you say we need to create the default indexes, then let's add those instructions.

en_us/install_operations/source/platform_releases/ginkgo.rst
```diff
@@ -0,0 +1,219 @@
...
+not normally use the ``guest_email`` column. If your installation does not use
+this column, you can avoid the table lock by using the ``--fake`` argument in
+migrating the ``orders`` table, running the migrate command in this form:
+
+.. code-block:: bash
+
+   ./manage.py migrate orders 0013 --fake
+
+=====================
+Upgrading Celery
```
grantgoodmanedX: @nedbat Can we drop this section? I think it's covered by the migration instructions above.

nedbat: Oh, right. I meant to do that, sorry.  Drop this section.

catong:
@grantgoodmanedX Looks good; I built each Gingko version guide included in this PR and each one built successfully, looked fine, but 2 of the guides have unrelated build errors that should clear after you rebase against master. 👍

PR #1569, opened 2017-08-14
Ned tweaks for the Ginkgo instructions.

nedbat:


nedbat:
@grantgoodmanedX LOOK :)

PR #1617, opened 2017-11-03
Mobile doc: Support for Android tablets too

catong:
## [DOC-3811](https://openedx.atlassian.net/browse/DOC-3811)

Update SFD mobile docs to indicate support on tablets both Android and iOS.

### Reviewers

- [x] Doc team review: @edx/doc
- [ ] Product review: @marcotuts 

### Testing

### Post-review

- [x] Squash commits

### DIFF
--- a/en_us/shared/students/SFD_mobile.rst
+++ b/en_us/shared/students/SFD_mobile.rst
@@ -4,8 +4,8 @@
 If You Use the edX Mobile App
 ###############################
 
-This topic answers questions about how you can use the edX mobile app on an
-Android smartphone or an iPhone to take edX courses.
+This topic answers questions about how you can use the edX mobile app on a
+smartphone or tablet to take edX courses.
 
 .. contents::
   :local:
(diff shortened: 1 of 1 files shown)


en_us/shared/students/SFD_mobile.rst
```diff
@@ -65,10 +65,9 @@ You can use the app to do the following things.
  What platforms does the app run on?
  ========================================
 
- Right now, the edX mobile app is available for Android smartphones, iPhones,
- and the iPad. For the iPhone app, your device must be running iOS 7 or later.
-
- EdX plans to make the app available on Android tablets in the future.
+ The edX mobile app is available for smartphones and tablets on both Android
```
debchatigny: Is edX mobile app a productized name? If so, it should be edX Mobile App. Otherwise, should it be:
"The edX app (or just app) is available for smartphones and tablets ..." I see more contemporary instances where 'mobile' is not used, and realize this may be a larger question around product naming/Marketing.

debchatigny:
I just have the one inline comment around the name of the app. Otherwise, looks great!

en_us/shared/students/SFD_mobile.rst
catong: It's not a productized name. Good point, no need to say "mobile".

catong:
Merging https://github.com/edx/edx-documentation/pull/1617

PR #1640, opened 2018-01-12
DOC-3867: Update Python style guideline to allow "Args".

robrap:
## [DOC-3867](https://openedx.atlassian.net/browse/DOC-3867)

Update Python style guideline to allow "Args" in docstrings.

Some additional thoughts I had while looking at that guidelines fall into two categories:
- Guidelines that must be followed.
  - For these, we generally rely on automation to break a build.
  - There are places where the guideline makes it seem like it is required, but we don't break the build.
    - In this case, is it really just a recommendation? 
- Guidelines that are recommendations.
  - How should reviewers handle these?  Either reviewers can:
    - manually raise awareness with a "Nit" comment for every violation (BAD IDEA), or
    - manually raise awareness at random with a "Nit" comment, or
    - is it possible to produce automated warnings to a separate report that someone could read to learn these recommendations?
    - It would be great if we documented the answers to this as well.
  - How should our documentation handle these? Should it make it clear it is a recommendation vs a rule?
    - If it is worded like a required rule, anal reviewers like myself don't know what to do.
    - If it is clearly stated to be just a recommendation, are we fine if is ignored by coder and/or reviewer?

An example in this document is:
```
Not like this::

    def foo(a, b):
        """Computes the foo of a and b.""" # NO NO NO
```
Many people still do this.  I'm pretty certain it doesn't break the build.  Do we actually care at all about this if it isn't in our automated rules?

### Reviewers

- [x] Subject matter expert: @nedbat  
- [ ] Subject matter expert: @cpennington 

FYI: @nasthagiri 

### Testing

### HTML Version (optional)

### Sandbox (optional)

### Post-review

- [ ] Squash commits

### DIFF
--- a/en_us/developers/source/style_guides/python-guidelines.rst
+++ b/en_us/developers/source/style_guides/python-guidelines.rst
@@ -199,7 +199,7 @@ Here's how you write documentation in a mostly "Google Style" manner::
 
   * The summary line is on the second line, including single-line comments
     (see below)
-  * Use the full word "Arguments".
+  * Use the full word "Arguments", although "Args" is also acceptable.
 
 Most of our code is written using an older style::
 


nedbat (approved):

robrap:
@grantgoodmanedX: This is ready to merge whenever you are ready.  Thanks.

grantgoodmanedX (approved):
👍  Looks good.

PR #1642, opened 2018-01-12
DOC-3860 Add Course Sessions info to Learner's Guide

catong:
## [DOC-3860](https://openedx.atlassian.net/browse/DOC-3860)

Update LG for bundling v2 (courses in programs have multiple sessions)

### Reviewers

- [x] Subject matter expert: @Mjloturco 
- [ ] Subject matter expert: @mjfrey 
- [x] Doc team review (sanity check, copy edit): @edx/doc 
- [x] Product review: @dabdul-hathi

FYI: @dhixon 

### Testing

### Draft HTML Version 
- [x] http://draft-bundling-v2-lg.readthedocs.io/en/latest/SFD_enrolling.html#enroll-in-a-program
- [x] http://draft-bundling-v2-lg.readthedocs.io/en/latest/SFD_enrolling.html#course-sessions

### Post-review
- [x] Squash commits


Mjloturco (approved):
I believe that the time-frames for refund/select session may be different depending on the enrollment mode (particularly between verified and professional), I'm not sure if we should mention that or exactly what the differences are (@dabdul-hathi may have more details)

Otherwise looks great!

catong:
Thanks @Mjloturco You are right. I will try to work those details in without making it too confusing.

en_us/students/source/SFD_enrolling.rst
```diff
@@ -82,6 +82,45 @@ For more information, see :ref:`course start dates<Start Date and Duration>`
 and :ref:`course pacing<SFD Course Pacing>`.
 
 
+.. _Course Sessions:
+
+============================
+Course Sessions
+============================
+
+In some programs, courses offer multiple sessions and the ability to change
```
debchatigny: In 5.1, we say learners need to understand several concepts, but don't define them here. In 5.1.2, we don't really define what a **Course Session** is either; I think starting this section with a one-line definition of what a course session is would be very helpful.

en_us/students/source/SFD_enrolling.rst
```diff
@@ -82,6 +82,45 @@ For more information, see :ref:`course start dates<Start Date and Duration>`
...
+between sessions. For these courses, on the **Courses** and **Programs** pages
+of your dashboard, after you enroll in the course, you have options to select a
+session, change to a different session, or leave a session. You must select a
+session before you can access the course.
+
+For courses that offer multiple sessions, be aware of the following timeframes.
+
+* If you were already enrolled in the audit track of a course and you purchase
```
debchatigny: This comment applies to all of the following bullets. I feel like this information would work better in a table format, as all of the dates/timeframes are buried within the text of the bullets. Maybe next week we can brainstorm a table format for this information.

Content/structure looks fine, but I just don't believe it's delivered in the most scannable/usable format in bullets.

debchatigny:
Hi Carol - I added a few notes (suggestions) inline. Otherwise, all looks fine!

en_us/students/source/SFD_enrolling.rst
```diff
@@ -82,6 +82,45 @@ For more information, see :ref:`course start dates<Start Date and Duration>`
...
+  the verified track as part of a program purchase, we automatically enroll you
+  in the verified track for the course, in the current session. You can change
+  to a different session if the one we enrolled you in does not work for you.
+
+* After you join a course session (including if we automatically add you to a
+  session as a result of your course purchase), you can change to another
+  session or leave the session and wait to join a later session, within 14 days
```
dabdul-hathi: as Turc mentioned the 14 day policy is for verified track, and prof ed will be 2 days. Majority of programs will be verified track, so we should find a way to put more weight on that, but still provide info about the prof ed courses.

dabdul-hathi: should also add that - they can switch as long as they haven't completed the course.

dabdul-hathi: wondering if we should add the fact that their progress is not carried through from one session to another session.

en_us/students/source/SFD_enrolling.rst
catong: The local TOC directive under the intro paragraph generates a list of subtopic links. The rendered HTML looks like this: 
enrollingconcepts

catong: Agree that it would be good to define course session.

en_us/students/source/SFD_enrolling.rst
catong: Agree that a table would help the presentation, but I believe that this content will migrate to the Help Center fairly soon? This topic currently is very similar to the Help Center article. Presenting the information differently *might* cause some confusion if it takes readers time to realize they have already the same info in the HC? We should follow up about HC migration plans.

catong:
@dabdul-hathi @debchatigny Made updates based on your reviews. Please take a look at the regenerated draft HTML.

catong:
Merging https://github.com/edx/edx-documentation/pull/1642

PR #1646, opened 2018-01-25
Research Guide: Update description of 'location' in upgrade.clicked event

grantgoodmanedX:
## [DOC-3865](https://openedx.atlassian.net/browse/DOC-3865)

Changes to the description of the 'location' field, from https://openedx.atlassian.net/wiki/spaces/AN/pages/53510700/Course+Home+Page+Event+Design?focusedCommentId=471400694#comment-471400694

### Reviewers

- [x] Subject matter expert: @robrap 

### Testing

### Post-review

- [x] Squash commits

### DIFF
--- a/en_us/data/source/internal_data_formats/tracking_logs/student_event_types.rst
+++ b/en_us/data/source/internal_data_formats/tracking_logs/student_event_types.rst
@@ -2216,7 +2216,7 @@ event type also includes the following ``context`` member field.
    * - ``location``
      - string
      - The location in the LMS where the student clicked an upgrade link or
-       button: 'date-sidebar', 'hero', 'learner_dashboard', or 'sock'.
+       button.
    * - ``mode``
      - string
      - Enrollment mode when the user clicked an upgrade link or button:


en_us/data/source/internal_data_formats/tracking_logs/student_event_types.rst
```diff
@@ -2216,7 +2216,9 @@ event type also includes the following ``context`` member field.
    * - ``location``
      - string
      - The location in the LMS where the student clicked an upgrade link or
-       button: 'date-sidebar', 'hero', 'learner_dashboard', or 'sock'.
+       button: 'sidebar-message', 'learner_dashboard', or 'sock'. The location
```
robrap: This is definitely more accurate. Thank you.
Another possibility would be to have it end with `... or button.`, and not provide the detailed values so that they can change over time without having to update the docs.
@stroilova @marcotuts ?

stroilova: @robrap  I agree with your instinct & direction. Especially with these upgrade events that I think were not even intended for research originally, it seems like we'd want to minimize the number of potential future updates to eventing and docs. I am fine with removing the values. We can add them back in if really needed.

I had thought these were BI events that we accidentally included, and that we are now maintaining... If they get too complicated for us to maintain, we might want to considerdeprecating them from the research data package, moving them to the BI namespace & only storing internally. cc @robrap @marcotuts @

grantgoodmanedX: @robrap I've made the suggested change.

robrap (approved):

srpearce:
👍

PR #1654, opened 2018-02-08
Removing release notes from the build-rtd-documents script

srpearce:
We don't create new release notes anymore, and changes to documentation can cause errors in the old release notes. Removing the release notes from the automatic build script.


srpearce:
@grantgoodmanedX is taking care of this in a separate PR.

PR #1656, opened 2018-02-08
Removing release notes from build-rtd-documents script

grantgoodmanedX:
We don't create new release notes anymore, and changes to documentation can cause errors in the old release notes. Removing the release notes from the automatic build script.

### Reviewers

### Testing

### Post-review

- [x] Squash commits


catong (changes requested):
Can we also remove release notes from run_tests.sh while we're at it?
Possibly also the top level (en_us) Makefile.

catong:
👍 after squashing commits

PR #1682, opened 2018-04-02
Giving reason and role when enrolling manually on Instructor Dashboard

grantgoodmanedX:
## [DOC-3903](https://openedx.atlassian.net/browse/DOC-3903)

In the sections about enrolling and unenrolling learners from the Instructor Dashboard, I added notes about the new Role and Reason fields.

### Date Needed

The feature has already been released, so it would be good to get the docs published promptly.

### Reviewers

- [x] Subject matter expert: @bill-filler 
- [ ] Subject matter expert: 
- [ ] Product review:
- [ ] Partner support: 
- [ ] PM review: 

### Testing

### Post-review

- [ ] Squash commits

### DIFF
--- a/en_us/shared/manage_live_course/course_enrollment.rst
+++ b/en_us/shared/manage_live_course/course_enrollment.rst
@@ -123,6 +123,20 @@ To enroll learners or course team members, follow these steps.
    note that this feature is better suited to courses with smaller enrollments,
    rather than courses with massive enrollments.
 
+#. For **Role of the users being enrolled**, select the role of the learner.
+
+   * If the learner is a member of the course staff, select **Partner**.
+   * If the learner is not a member of the course staff, select **Learner**.
+
+   .. note::
+    All of the users that you enroll at one time must have the same role. If
+    you have some users who are partners and others who are learners, you must
+    complete two batch enrollments.
+
+#. In the **Enter the reason why the students are to be manually enrolled or
+   unenrolled** field, enter a specific, detailed reason why you want to
+   enroll these learners.
+
 #. To streamline the course enrollment process, leave **Auto Enroll** selected.
 
 #. To send learners an email message, leave **Notify students by email**
@@ -208,6 +222,10 @@ To unenroll learners, you supply the email addresses of enrolled learners.
 #. In the **Batch Enrollment** section of the page, enter a username or an
    email address, or multiple names or addresses separated by commas or new
    lines.
+
+#. In the **Enter the reason why the students are to be manually enrolled or
+   unenrolled** field, enter a specific, detailed reason why you want to
+   unenroll these learners.
 
 #. To send learners an email message, leave **Notify students by email**
    selected.


bill-filler (approved):

en_us/shared/manage_live_course/course_enrollment.rst
```diff
@@ -123,6 +123,12 @@ To enroll learners or course team members, follow these steps.
    note that this feature is better suited to courses with smaller enrollments,
    rather than courses with massive enrollments.
 
+#. Select the role of the learner. Select **Partner**, for instance, if the
```
srpearce: Even though this UI label isn't very good, I think we need to indicate where the user performs this action. I also think we can make the language stronger, and move the info out of paragraph form into bullet form to make it more visible. And it looks like you can't add partner users and learner users at the same time. So I'd change this step to:

For **Role of the users being enrolled**, select the role of the learner.
* If the learner is a member of the course staff, select **Partner**.
* If the learner is not a member of the course staff, select **Learner**.
Note 
All of the users that you enroll at one time must have the same role. If you have some users who are partners and others who are learners, you must complete two batch enrollments.

I'd also suggest reviewing this UI text and creating a platform PR to update it, since it didn't have a doc review before. That would mean we'd have to update these docs later, but the changes would be small. For example, I'd change "Role of the users being enrolled" to "Role for these users" or similar.

en_us/shared/manage_live_course/course_enrollment.rst
```diff
@@ -123,6 +123,12 @@ To enroll learners or course team members, follow these steps.
...
+   learner is a member of the course staff. Select **Learner** if the learner
+   is not a member of the course staff.
+
+#. Enter the reason that you are enrolling the learners manually.
```
srpearce: Again, even though this label is terrible (which could very well be an a11y issue), we should tell the user where to perform the action:

In the **Enter the reason why the students are to be manually enrolled or unenrolled** field, enter a specific, detailed reason why you want to enroll these learners.

For the current UI text (Enter the reason why the students are to be manually enrolled or unenrolled. This cannot be left blank and will be recorded and presented in Enrollment Reports. Therefore, please give enough detail to account for this action.), I have some suggestions.
* Name the field "Reason for Enrollment or Unenrollment". 
* Use an edited version of the current three sentences as help text.
* Remove the "Reason" placeholder from the field. (Per Darren in UX, placeholder text creates a11y and other problems.)

en_us/shared/manage_live_course/course_enrollment.rst
```diff
@@ -209,6 +215,8 @@ To unenroll learners, you supply the email addresses of enrolled learners.
    email address, or multiple names or addresses separated by commas or new
    lines.
 
+#. Enter the reason that you are unenrolling the learners.
```
srpearce: In the **Enter the reason why the students are to be manually enrolled or unenrolled** field, enter a specific, detailed reason why you want to unenroll these learners.

srpearce (changes requested):

grantgoodmanedX:
@srpearce I've updated this with your suggested changes.

en_us/shared/manage_live_course/course_enrollment.rst
```diff
@@ -123,6 +123,20 @@ To enroll learners or course team members, follow these steps.
    note that this feature is better suited to courses with smaller enrollments,
    rather than courses with massive enrollments.
 
+#. For **Role of the users being enrolled**, select the role of the learner.
+
+   * If the learner is a member of the course staff, select Partner.
+   * If the learner is not a member of the course staff, select Learner.
```
srpearce: UI terms should be bold in the above.

srpearce (approved):
Approved with one small change.

PR #1691, opened 2018-05-01
Deb/DOC 3834 Add info to LTI description fields

debchatigny:
Added info to LTI description fields on why LTI providers, requesting  usernames and emails, need to launch in a new window, so as to present a warning message to learners so that they’re aware they’re providing this information to an LTI provider/third-party. 

See [DOC-3834](https://openedx.atlassian.net/browse/DOC-3834) for details provided by Douglas Hall.


srpearce:
👍

PR #1693, opened 2018-05-07
TE-2567 Better requirements file management

jmbowman:
## [TE-2567](https://openedx.atlassian.net/browse/TE-2567)

This updates the management of Python package dependencies to comply with OEP-18, which should help make documentation builds more reliable over time and make it easier to upgrade dependencies like Sphinx when we choose.  There were originally two requirements files in the repository; one has been updated to just reference the new file in the standard location, and the other seems to be unused even in RTD project settings, so it's being removed.

I also updated one outdated link I found - `sphinx-napoleon` was merged into `sphinx` a while back.

### Date Needed

We would like this to be merged before the Hawthorn master branch is created, so Hawthorn doc builds are less likely to randomly fail over time.

### Reviewers

The PR submitter checks the boxes after each reviewer finishes and gives :+1:. 

- [ ] Subject matter expert: @nedbat

### Testing


grantgoodmanedX:
👍 This looks sensible to me. As you note, we'll need to coordinate to update all the existing RTD projects to use the proper requirements file.

Makefile
```diff
@@ -1,2 +1,22 @@
+# Do things in edx-platform
```
nedbat: This isn't edx-platform.

nedbat (changes requested):

Makefile
nedbat: Good to merge once fixed :)

nedbat (approved):

Makefile
jmbowman: Fixed the copy-pasta.

PR #1698, opened 2018-05-17
Fix formatting for LaTeX info in HTML component topic

srpearce:
Fix the headings for the LaTeX info section in the HTML component topic.


debchatigny:
👍

PR #1699, opened 2018-05-17
Consolidate notes in HTML image topic

srpearce:
Consolidate two separate notes, each with two bullet points, into one note with four bullet points for readability


debchatigny:
👍

PR #1725, opened 2018-08-01
Draft release note, based on Matt Dubose portal post

grantgoodmanedX:
Draft of release notes for Hawthorn, starting from @mduboseedx 's post at https://open.edx.org/blog/all-about-hawthorn

Need to verify that all these feature changes are in Open edX and not just edx.org
Also need to update System Upgrades and Updates section, which is currently just a cut and paste from Ginkgo and therefore 100% wrong.

### Date Needed (optional)

8/2

### Reviewers

Possible roles follow. The PR submitter checks the boxes after each reviewer finishes and gives :+1:. 

- [ ] Subject matter expert: @mduboseedx 
- [x] Subject matter expert: @nedbat 
- [x] Doc team review (sanity check, copy edit, or dev edit?): @edx/doc
- [ ] Product review:
- [ ] Partner support: 
- [ ] PM review: 

### Testing

- [x] Ran ./run_tests.sh without warnings or errors

### Post-review

- [ ] Squash commits

PR #1726, opened 2018-08-03
Set Open edX variables before pulling repos, etc.

nedbat:
## [DOC-XXXX](https://openedx.atlassian.net/browse/DOC-XXXX)

The old way, big things would get done before the Open edX variables were defined, which was bad.
I'm not sure if I removed too much though.  @jmbowman: correct me.

### Date Needed (optional)

If the release date of a feature is known or estimated, provide it to give reviewers guidance on turnaround time.

### Reviewers

- [ ] Subject matter expert: @jmbowman 
- [ ] Subject matter expert: 
- [ ] Product review:
- [ ] Partner support: 
- [ ] PM review: 

### Testing

### HTML Version (optional)

### Sandbox (optional)

### Post-review

- [ ] Squash commits

### DIFF
--- a/en_us/install_operations/source/installation/install_devstack.rst
+++ b/en_us/install_operations/source/installation/install_devstack.rst
@@ -29,17 +29,38 @@ Install Devstack
 
 To install Devstack, follow these steps.
 
-#. Check out a local copy of the ``edx/devstack`` repository from
-   ``https://github.com/edx/devstack``, selecting the branch that you want to
-   work with.
+#. Decide which branch you will be working with.  "master" is the latest code
+   in the repositories, changed daily.  Open edX releases are more stable,
+   for example, Hawthorn.
 
-#. Navigate to the ``devstack`` directory.
+#. Check out a local copy of the ``edx/devstack`` repository from
+   ``https://github.com/edx/devstack``.
+
+    .. code-block:: bash
+
+     git clone https://github.com/edx/devstack
+     cd devstack
+
+#. If you are not using the master branch, check out the branch you want.
+
+    .. code-block:: bash
+
+     git checkout open-release/hawthorn.master
+
+#. If you are not using the master branch, define an environment variable for
+   the Open edX version you are using, such as ``hawthorn.master`` or
+   ``zebrawood.rc1``. Note that unlike a server install, the value of the
+   ``OPENEDX_RELEASE`` variable should not use the ``open-release/`` prefix.
+
+    .. code-block:: bash
+
+     export OPENEDX_RELEASE=hawthorn.master
 
 #. Create a Python virtual environment. For information about how to do this,
    see `Virtual Environments`_.
 
 #. Install the requirements inside the Python virtual environment.
-   
+
     .. code-block:: bash
 
      make requirements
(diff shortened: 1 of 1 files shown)


grantgoodmanedX:
The sphinx warnings are fixed, I believe, in my PR https://github.com/edx/edx-documentation/pull/1725/

en_us/install_operations/source/installation/install_devstack.rst
```diff
@@ -29,12 +29,33 @@ Install Devstack
...
+    .. code-block:: bash
+
+     git checkout open-release/hawthorn.master
+
+#. If you are not using the master branch, define an environment variable for
+   the Open edX version you are using, such as ``hawthorn.master`` or
+   ``zebrawood.rc1``. Note that unlike a server install, the value of the
+   ``OPENEDX_RELEASE`` variable should not use the ``open-release/`` prefix.
 
-#. Navigate to the ``devstack`` directory
```
jmbowman: It seems like we'd want to keep this step for the following ones involving `make` commands.

en_us/install_operations/source/installation/install_devstack.rst
```diff
@@ -104,24 +125,6 @@ following sequence of commands.
...
-To run an image of an Open edX release other than the latest, follow these 
-steps.
-
-#. Set the ``OPENEDX_RELEASE`` environment variable to the appropriate tag 
-   for the image tag that you want to run, such as ``hawthorn.master``
-   or ``zebrawood.rc1``. Note that unlike a server install, the value of the 
-   ``OPENEDX_RELEASE`` variable should not use the ``open-release/`` prefix.
-
-#. Run ``make dev.checkout`` to check out the correct branch in the local 
-   checkout of each service repository.
```
jmbowman, line 119: This step needs to be preserved in the updated instructions above (it's optional when using `master`).

en_us/install_operations/source/installation/install_devstack.rst
nedbat: I included the cd in the clone step, so that the git checkout would work.  We can put it back to a separate step I guess.

nedbat:
@jmbowman Thanks for the review.  I've restored the make dev.checkout command.  Can you take another quick look to see that it's in the right place? Thanks.

en_us/install_operations/source/installation/install_devstack.rst
```diff
@@ -29,17 +29,50 @@ Install Devstack
...
+
 #. Create a Python virtual environment. For information about how to do this,
    see `Virtual Environments`_.
 
+#. Run ``make dev.checkout`` to check out the correct branch in the local
+   checkout of each service repository.
+
+    .. code-block:: bash
+
+     make dev.checkout
```
jmbowman, line 69: It seems odd to have this between creating the virtualenv and installing requirements in it, but also...I don't think we need the virtualenv anymore.  https://github.com/edx/edx-documentation/blob/master/en_us/install_operations/source/installation/installation_prerequisites.rst#software-prerequisites covers needing `docker-compose` installed, and that's the only thing we used this virtualenv for.  So I think you're ok just removing the preceding and following steps.  Otherwise, 👍

PR #1729, opened 2018-08-20
Fix JSON format text-wrapping issue.

debchatigny:
In file 17.2.1.4, fix JSON text formatting that is running off the page.

I fixed two issues:
 - Added line breaks to the end of all lines in the code block.
 - Added grave double accents (``) at the beginning and the end of the
lines.

https://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/s
tudent_progress/course_answers.html?highlight=answer%20data#interpret-th
e-student-state-report

### Date Needed: 21-Aug-2018

If the release date of a feature is known or estimated, provide it to give reviewers guidance on turnaround time.

### Reviewers

- [ ] Subject matter expert: 
- [ ] Subject matter expert: 
- [ ] Product review:
- [ ] Partner support: 
- [ ] PM review: 

### Testing

### HTML Version (optional)

### Sandbox (optional)

### Post-review

- [ ] Squash commits

### DIFF
--- a/en_us/shared/student_progress/course_answers.rst
+++ b/en_us/shared/student_progress/course_answers.rst
@@ -256,7 +256,12 @@ When you open the report, the value in the **State** column appears on a single
 line. This value is a record in JSON format. An example record for a text input
 CAPA problem follows.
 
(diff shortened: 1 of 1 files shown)


grantgoodmanedX:
I don't think this change is necessary. The point of the example at line 259 is to show hard-to-read JSON, all on one line, which is the default output in the Student State report, and demonstrate how you can use a tool to make it more readable in the example at line 264.

debchatigny:
This was the file that shows the JSON code spilling off of the screen, so there's no way to know how much JSON text there is. While I agree that it's just an example, the example doesn't show the end result as it is. You can view the unedited file here:

https://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/student_progress/course_answers.html#interpret-the-student-state-report

How else would you suggest to fix this, @grantgoodmanedX?

grantgoodmanedX:
I think I would leave it alone. The text says "When you open the report, the value in the State column appears on a single line." and the example also shows the value on a single line. The text goes on to explain (as you and the user have noticed) that this output isn't very readable, but the user can use tools to re-format it into a readable form. Changing the example into a readable form would miss the point, I think.

PR #1821, opened 2019-04-30
Adds django linter violations to xss linter.

adeelkhan:
This adds documentation for django linter violations that is added in https://github.com/edx/edx-platform/pull/19615

LEARNER-4632

### DIFF
--- a/en_us/developers/source/preventing_xss/preventing_xss.rst
+++ b/en_us/developers/source/preventing_xss/preventing_xss.rst
@@ -1355,6 +1355,186 @@ Once the page level directive has been added to make the Mako template safe by
 default, any use of the ``h`` filter in an expression is redundant. These ``h``
 filters should be removed.
 
+django-trans-missing-escape
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Each translation string needs to be properly escaped before being shown to the user. This error is raised when a ``trans`` tag has a missing escape filter. Translation strings should be passed through django's builtin ``force_escape`` filter.
+
+.. code::
+
+    ## DO this
+    {% trans 'Log out' as tmsg %}{{tmsg|force_escape}}
+
+The ``force_escape`` filter should be on the same line as the ``trans`` expression and should not be moved to next line.
+
+.. code::
+
+    ## DO NOT do this
+    {% trans 'Log out' as tmsg %}
+    {{tmsg|force_escape}}
+
+django-trans-invalid-escape-filter
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+This error is raised when the ``trans`` tag is escaped via unknown filter.
+
+.. code::
+
+    ## DO Not do this
+    {% trans 'Log out' as tmsg %}{{tmsg|some_unknown_filter}}
+
+
+django-trans-escape-variable-mismatch
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+This error is raised when there is a mismatch between the ``trans`` tag expression variable and the filter expression variable.
+
+.. code::
+
+    ## DO Not do this
+    {% trans 'Log out' as tmsg %}{{some_other_variable|force_escape}}
+
+
+django-html-interpolation-missing
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Whenever there are some html tags that need to be used in the translation string and we do not want them to be escaped then we have to interpolate such html via a custom ``interpolate_html`` tag.
+
+This error is raised when a trans/blocktrans tag includes html directly, rather than using the ``interpolation_html`` tag.
+
+In the case of a ``trans`` tag::
+
+    ## DO NOT do this
+    {% trans "some text <a href='some-path'>link text to display</a>." %}
+
+The correct way is to use a variable to store the translation string and use the ``interpolate_html`` tag to escape that variable like below.
+
+.. code::
+
+    ## DO this
+    {% trans "Some text {start_link}link text to display{end_link}." as tmsg %}
+    {% interpolate_html tmsg start_link='<a href='some-path'>'|safe end_link='</a>'|safe %}
+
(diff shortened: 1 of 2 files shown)


en_us/developers/source/preventing_xss/preventing_xss.rst
```diff
@@ -1355,6 +1355,186 @@ Once the page level directive has been added to make the Mako template safe by
...
+    ## DO NOT do this
+    {% blocktrans%}
+    some text <a href="some-path">link text to display</a>.
+    {% endblocktrans %}
+
+Instead we should use the ``interplate_html`` tag to HTML-escape the string, except for parts that are real HTML that are marked as ``safe``.
+
+.. code::
+
+    ## DO this
```
robrap, line 1425: @adeelkhan: What would happen if any of these that use `interpolate_html ` were mistakenly also wrapped with `{% filter force_escape %}`?  Will the linter find that and complain?  Is it a problem?

robrap (commented):
Thanks @adeelkhan.  This looked really good.  I figured it would be simpler to make my comments as changes.  Please review my changes and ensure I didn't introduce any typos or issues.

I made the following changes:
- I used `start_link` instead of `start_anchor`, etc., so the names make more sense to translators.
- I update spacing.
- I removed new lines in long text to match our newer spec for documentation.
- I update the examples with a url and switched to `some_url` to make it shorter.  Please ensure I did this correctly.
- Various other edits.

Thank you!

en_us/developers/source/preventing_xss/preventing_xss.rst
adeelkhan: For now linter would not complain, but i think this can be added.  I dont think it is correct to use both and any developer would use them simultaneously, but i think django template would not complain too. The question is we can use a tag inside a filter in django templates. I have to check that.

The idea is linter inspects between blocktrans/endblocktrans content to see if there is any html/interpolated variable inside it or not. Based on it decides
	1) If foudn regular string, move up and treat it as regular blocktrans tag and find filter expression for such block.
	2) If found any html content, flag it as violation, we would not allow this.
	3) If found any interpolation variable eg {var}, then look for connected interpolate_html expression. It connects them using output variable used like in below example variable is tmsgx

So there are three cases: 
```{% filter force_escape %}
  {% blocktrans trimmed asvar tmsgx %}
      {{span_start}} hour{% plural %}{{span_end}} hours
  {% interpolate_html tmsgx span_start='<span>'|safe span_end='</span>'|safe %}
  {% endblocktrans %}
{% endfilter %}
```
Linter would complain as it finds html tags inside blocktrans/endblocktrans, so it doesnt checks if its under filter/endfilter block. We would not allow html usage inside blocktrans/endblocktrans.

``` {% filter force_escape %}
  {% blocktrans trimmed asvar tmsgx %}
    {{span_start}} hour{% plural %}{{span_end}} hours
  {% endblocktrans %}
  {% interpolate_html tmsgx span_start='<span>'|safe span_end='</span>'|safe %}
{% endfilter %}
```
Now as there is no html/interpolate variable inside blocktrans/endblocktrans, linter would expect it as a regular block and move up to find {% filter force_escape %} expression.

``` {% filter force_escape %}
  {% blocktrans trimmed asvar tmsgx %}
      {span_start} hour{% plural %}{span_end} hours
  {% endblocktrans %}
  {% interpolate_html tmsgx span_start='<span>'|safe span_end='</span>'|safe %}
{% endfilter %}
```
Linter finding interpolation variable like {span_start} would try to find related interpolate_html and further checks it has the correct arguments and any html used is passed through safe filter.

There is no check on the position of interpolate_tag. I guess a blocktrans/endblocktrans could be in one position and its not necessary for its interpolate_html tag to be aligned with it. It can be anywhere in file.

en_us/developers/source/preventing_xss/preventing_xss.rst
```diff
@@ -1382,101 +1379,90 @@ This error is raised when the ``trans`` tag is escaped via unknown filter.
...
+In the case of a ``blocktrans` tag::
 
     ## DO NOT do this
     {% blocktrans%}
-    some text <a href="some-path"> anchor text for to display </a>.
+    some text <a href="some-path">link text to display</a>.
     {% endblocktrans %}
 
-Instead we should use ``interplate_html`` tag to escape string while keeping intact inner html.
+Instead we should use the ``interplate_html`` tag to HTML-escape the string, except for parts that are real HTML that are marked as ``safe``.
```
adeelkhan: interplate_html should be interpolate_html

en_us/developers/source/preventing_xss/preventing_xss.rst
```diff
@@ -1382,101 +1379,90 @@ This error is raised when the ``trans`` tag is escaped via unknown filter.
...
-            ``interpolate_html`` ie ``msg``, should map to ``trans`` tag output variable, otherwise linter would not
-            be able to connect to its related ``trans`` expression and would complain above error.
+.. Important:: One thing to keep in mind while using ``interpolate_html`` is to correctly map the ``trans/blocktrans`` tag string output variable to the ``interpolate_html`` tag.  For example, in the above case the ``msg`` variable used in ``interpolate_html`` must match the ``trans/blocktrans`` tag output variable (also ``msg``), otherwise the linter will complain.
 
 
 django-html-interpolation-missing-safe-filter
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-This error is raised when the injected html is not marked as safe via ``safe`` filter. Not marking it safe would cause
-``interpolate_html`` tag to not only escape the translation string but the arguments passed to it.
+This error is raised when the injected html is not marked as safe via a ``safe`` filter. Not marking it safe would cause the ``interpolate_html`` tag to not only HTML-escape the translation string, but it would also HTML-escape the the true HTML arguments that were passed to it.
```
adeelkhan: "the" comes twice

robrap (approved):
Thanks.

PR #1846, opened 2019-11-18
There's no such thing as the Student Engagement Report

nedbat:
Removed discussion of a feature that doesn't usefully exist.


nedbat:
@OmarIthawi @MHaton better to not have the page, right?

MHaton:
@nedbat Agreed, no feature - no doc ¯\\_(ツ)_/¯

PR #1848, opened 2020-01-16
Suggest smaller Python line-length limit

kdmccormick:
@nedbat 

I disagree that 120 characters is a good line-length limit.

I frequently code on my Thinkpad without an external monitor. Here's my screen when I edit a repo where we limit line length to 120. The vertical white line is 88 characters.

linelength-120

I can't see the end of every line without shrinking my terminals or zooming way out, both of which are annoying and not good for eye strain.

Furthermore, I don't have any serious vision impairments, so I can read fairly small text on a screen. Imagine if I did!

Here's the same bit of code, but Black-ed to wrap at column 88:
linelength-88_2

It really makes a difference!


timmc-edx:
The usual limiter for me is side-by-side diff views on a laptop, where I can get 80-100 columns in each pane before having to scroll horizontally or linewrap—either of which makes diffs very hard to read.

kdmccormick:
Closing (for now, at least).

PR #1858, opened 2020-03-31
Notifier Deprecation doc updates

sethmccann:
removed mention of notifier (i.e. daily digest, weekly email updates) and updated images associated Notifier which display the "send updates" button which is now removed.

## [PROD-1342](https://openedx.atlassian.net/browse/PROD-1342)
 - no associated DOC Jira ticket

Add a description of your changes with links to any relevant material.

First step of deprecating Notifier was to hide on edx.org and update docs. following steps are to follow OEP-21

### Reviewers

- [ ] Subject matter expert:  @bderusha 

### Testing

None

### Post-review

- [ ] Squash commits


bderusha (approved):
Not 100% sure about whether the unsubscribe via email is still relevant, but I don't think having a whole section just for the fact that an unsubscribe link exists is worth the trouble.

PR #1862, opened 2020-04-24
Update enterprise api docs for new IDA endpoints

bbaker6225:
Add in documentation listing the new endpoints that will be available
under v2 of the enterprise API in API gateway.

ENT-2578

### DIFF
--- a/en_us/enterprise_api/source/api_reference/reference.rst
+++ b/en_us/enterprise_api/source/api_reference/reference.rst
@@ -27,13 +27,13 @@ switched over to v2.
 
 * **/v2/enterprise-catalogs** - You can make GET calls to the
   ``/enterprise-catalogs`` endpoint to get a list of all the enterprise course
-  catalogs. Note that this is not currently filtered to show only those available to your
-  organization. That functionality will be added in an upcoming release.
+  catalogs that are available to your organization.
 
 * **/v2/enterprise-catalogs/{catalog_uuid}** - You can make GET calls to the
   ``/enterprise-catalogs/{catalog_uuid}`` endpoint to get a list of all the
-  content in a specified course catalog.
-  For details, see :ref:`enterprise_catalogs_catalogID Endpoint`.
+  active courses in a specified course catalog. Active courses are courses that
+  are currently open for enrollment or that will open for enrollment in the
+  future. For details, see :ref:`enterprise_catalogs_catalogID Endpoint`.
 
 ============
 V1 Endpoints
--- a/en_us/enterprise_api/source/api_reference/reference.rst
+++ b/en_us/enterprise_api/source/api_reference/reference.rst
@@ -17,6 +17,28 @@ Endpoints
 
 The following endpoints are available in the Enterprise API.
 
+============
+V2 Endpoints
+============
+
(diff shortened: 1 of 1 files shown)


en_us/enterprise_api/source/api_reference/reference.rst
```diff
@@ -17,6 +17,28 @@ Endpoints
...
+At the moment, these v2 endpoints are called in the same manner as the old endpoints, and simply
+use a new service to provide their data. Eventually many of the v1 endpoints will be deprecated and
+switched over to v2.
+
+* **/v2/enterprise-catalogs** - You can make GET calls to the
+  ``/enterprise-catalogs`` endpoint to get a list of all the enterprise course
+  catalogs. Note that this is not currently filtered to show only those available to your
+  organization. That functionality will be added in an upcoming release.
+
+* **/v2/enterprise-catalogs/{catalog_uuid}** - You can make GET calls to the
```
bbaker6225, line 32: @brittneyexline The docs for the `v1` version of this endpoint say that it filters the content metadata to only `active` courses. I'm guessing this refers to https://github.com/edx/edx-enterprise/blob/77b769b0918cf6acdcb498c5804c733bb20b4279/enterprise/models.py#L1506 so I think we'll need to add a ticket to add that functionality to: https://github.com/edx/enterprise-catalog/blob/669fb768a85e4a74239abbc61b2a56802539629f/enterprise_catalog/apps/catalog/models.py#L338-L353

bbaker6225:
This should be ready for review now that the new endpoints are in API gateway

en_us/enterprise_api/source/api_reference/reference.rst
bbaker6225: Put up https://github.com/edx/enterprise-catalog/pull/91 to add that functionality

en_us/enterprise_api/source/api_reference/reference.rst
```diff
@@ -17,6 +17,28 @@ Endpoints
...
+V2 Endpoints
+============
+
+At the moment, these v2 endpoints are called in the same manner as the old endpoints, and simply
+use a new service to provide their data. Eventually many of the v1 endpoints will be deprecated and
+switched over to v2.
+
+* **/v2/enterprise-catalogs** - You can make GET calls to the
+  ``/enterprise-catalogs`` endpoint to get a list of all the enterprise course
+  catalogs. Note that this is not currently filtered to show only those available to your
```
brittneyexline: I'm thinking we should make sure that these endpoints are ready to be transitioned to before we publish their existence. That means addressing this filtering issue as well as any current bugs we need to fix around the get_content_metadata endpoints. As such, I propose that we update this to be what we expect it to be when all gaps are addressed and then merge this PR after that. Perhaps this can be part of the release plan ticket.

bbaker6225: Sure, I'll add that to the ticket

bbaker6225: Added to https://openedx.atlassian.net/browse/ENT-2716

PR #1890, opened 2020-10-22
Fix all broken references

nedbat:
(There is a Jira ticket somewhere about this, but I cannot find it. It's not in the DOC- project.)

When we removed the Learner book in June, we kept all the pages, but made them have only a single sentence about the learner book moving to Zendesk.  All references to those pages were broken.  This pull request restores the title of each page, and all the internal reference points, so that page references can resolve.

There were also three other broken references which I fixed.  There are now zero errors and zero warnings when building pull requests.


shared/conf.py
```diff
@@ -153,15 +153,14 @@ def feedback_form_url(project, page):
     'xblocktutorial' : ('https://edx.readthedocs.io/projects/xblock-tutorial/en/latest/', None),
     'installation' : ('https://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/', None),
     'olx' : ('https://edx.readthedocs.io/projects/edx-open-learning-xml/en/latest/', None),
-    'learners' : ('https://edx.readthedocs.io/projects/edx-guide-for-students/en/latest/', None),
+    'learners' : ('https://edx.readthedocs.io/projects/edx-guide-for-students/en/nedbat-fix-learner-links/', None),
```
jmbowman, line 156: I presume this is getting changed back before you merge?

jmbowman (approved):

shared/conf.py
nedbat: I was wondering if you would notice! I could change it before merge, or merge without deleting the branch, then make another PR to change it, then delete both branches, which is more complicated, but would produce all green builds.

PR #1892, opened 2020-11-25
Clean up references to deprecated edx-platform import paths

kdmccormick:
Code within edx-platform djangoapps should be referenced
via its full path. For example,

>    `track.tests`

should now be referenced as:

>    `common.djangoapps.track.tests`.

The old import paths are deprecated.

This is part of the import-shims cleanup, which resulted from the sys.path-hack removal effort. See the [forum post](https://discuss.openedx.org/t/koa-will-change-how-edx-platform-code-is-imported/3610) for details.

I'm keeping track of the list of these PRs on this edx-platform "parent PR": https://github.com/edx/edx-platform/pull/25684

### DIFF
--- a/en_us/developers/source/analytics.rst
+++ b/en_us/developers/source/analytics.rst
@@ -193,7 +193,7 @@ Here is an example of a subclass.
 
 ::
 
-    from track.tests import EventTrackingTestCase
+    from common.djangoapps.track.tests import EventTrackingTestCase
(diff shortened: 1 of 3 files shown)


kdmccormick:
@nedbat - I think this should be included in Koa.

nedbat (approved):

nedbat:
I've cherry-picked this to Koa.

PR #1935, opened 2021-04-16
docs: new Lilac pages

nedbat:
These are the Lilac-specific pages, following the process here: https://openedx.atlassian.net/wiki/spaces/COMM/pages/19662426/Process+to+Create+an+Open+edX+Release

### DIFF
--- a/en_us/open_edx_release_notes/source/lilac.rst
+++ b/en_us/open_edx_release_notes/source/lilac.rst
@@ -1,0 +1,10 @@
+.. _Open edX Lilac Release:
+
+######################
+Open edX Lilac Release
+######################
+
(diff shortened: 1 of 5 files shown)


en_us/open_edx_release_notes/source/lilac.rst
```diff
@@ -0,0 +1,10 @@
+.. _Open edX Lilac Release:
+
+######################
+Open edX Lilac Release
+######################
+
+This page will list the highlights of the Lilac release of the Open edX platform.
```
pdpinch, line 7: Is this the page that the release notes will go into?

nedbat: Yes!

PR #1950, opened 2021-06-03
docs: adding License

jinder1s:
All public github repositories should have a license. This adds the appropriate license for this repository.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,350 @@
+Creative Commons Attribution-ShareAlike 4.0 International Public
+License
+
+By exercising the Licensed Rights (defined below), You accept and agree
+to be bound by the terms and conditions of this Creative Commons
+Attribution-ShareAlike 4.0 International Public License ("Public
+License"). To the extent this Public License may be interpreted as a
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,438 @@
+Attribution-NonCommercial-ShareAlike 4.0 International
```
timmc-edx: I don't think we want the NC variant -- the existing license information in the footer is by-sa, not by-nc-sa.

timmc-edx: Hmm... maybe that's actually backed by the edx/docs.edx.org repo? But I'd be surprised if we wanted to use different licenses for these two.

jinder1s: that information comes from edx sphinx theme

jinder1s: Good point on np by-sa vs by-nc-sa

jinder1s: I'll change these with by-sa

jinder1s: Okay, changed to proper license

timmc-edx (approved):

PR #1979, opened 2021-09-07
Add Reusable Rubrics feature documentation

nsprenkle:
## [AU-284](https://openedx.atlassian.net/browse/AU-284)

Adds documentation for how to reuse rubrics across multiple ORAs.

### Date Needed (optional)

Roll-out with [AU-205](https://openedx.atlassian.net/browse/AU-205)

### Reviewers

Possible roles follow. The PR submitter checks the boxes after each reviewer finishes and gives :+1:. 

- [x] Subject matter expert: 
- [ ] Doc team review (sanity check): @edx/doc
- [x] PM review: @sapanathomas523 

FYI: @edx/masters-devs-gta 

### Testing

- [x] Ran ./run_tests.sh without warnings or errors

### Post-review

- [ ] Add a comment with the description of this change or link this PR to the next release notes task.
- [ ] Squash commits

PR #1984, opened 2021-10-01
make rubric notes "notes"

jansenk:
Small fix for https://github.com/edx/edx-documentation/pull/1979. Rather than just the text, "note", we should use that actual note macro


mattcarter (approved):
approved

PR #1990, opened 2021-11-03
fix: docs links fixing and removed unreleased feature

mehaknasir:
## [DOC-XXXX](https://openedx.atlassian.net/browse/DOC-XXXX)

TNL-TICKET: https://openedx.atlassian.net/browse/TNL-9207

### Reviewers

- [ ] Subject matter expert: 
- [ ] Subject matter expert: 
- [ ] Product review:
- [ ] Partner support: 
- [ ] PM review: 

@marcotuts @awaisdar001 

### Testing
 Ran ./run_tests.sh without warnings or errors

### Post-review

- [ ] Squash commits


en_us/shared/course_assets/pages.rst
```diff
@@ -290,7 +291,7 @@ Course teams can specify one or multiple times during which the discussion forum
 Configuring the Wiki application
 ================================
 
-For instructions on how to enable the wiki application see 11.3.2.3 Enabling the Wiki Application.
+For instructions on how to enable the wiki application see :ref:`11.3.2.3 Enabling the Wiki Application<Enable Wiki>`.
```
nedbat, line 294: When making references like this, you don't need to repeat the title.  You could use this, and Sphinx will pull in the title from the destination page:
```
:ref:`Enable Wiki`
```

en_us/shared/course_assets/pages.rst
```diff
@@ -98,8 +99,8 @@ To disable or enable the Teams application, follow these steps.
 
 #. Select **Apply** to save your configuration changes.
 
-Additional configuration for the Teams application can be found at 11.3.3.3
-Configuring the Teams application
+Additional configuration for the Teams application can be found at :ref:`11.3.3.3 Configuring the Teams application<Teams Configuration>`
```
marcotuts, line 102: The link would ideally be the exact name and number of the section, not "Teams Configuration" am I reading the markup correctly here?

marcotuts: Ah I just saw Ned's comment further down looks like a shorthand for Shinx

en_us/shared/course_features/teams/teams_setup.rst
```diff
@@ -99,27 +99,10 @@ These names and descriptions are visible to learners in the Teams application as
    teams associated with the deleted topic.
 
 
-**Types of Team Groups**
```
marcotuts, line 102: @mehaknasir I'm sorry, I missed the note that we had here  letting people know why the settings might not be visible. 

 note:: If you do not see all team group type options, check with your platform administrator
   to see if the relevant teams application features have been enabled.
   
I'm debating whether or not to keep this section in after all I will have an answer today for this, instead of having to go back and add this.

marcotuts: This section no longer needs to be removed given our existing note, this will give us open edx valid documentation with a way for edx partners to reach out if they have questions on this specific feature.

marcotuts (approved):
Other than keeping the team types section (a change from the original ticket's ACs) this is good to merge. thanks! 

edit: Only change missing is the Team Groups change included in the comment below. Ok to remove types of team groups section (https://github.com/edx/edx-documentation/pull/1990/files#issuecomment-961305855)

marcotuts:
From internal edX feedback, an additional requested change is in the Team Groups section

**Team Groups:**
Each team _group_ is a grouping for multiple learner teams, **and each group has several
configuration details and options.** To start, each of these team groups must be given a
unique name, and optionally a description can also be set.

The bolded section can be removed. The italicized word should become group not groups. 

FYI @sapanathomas523

en_us/shared/course_features/teams/teams_setup.rst
marcotuts: Sorry for the churn here. Ok after all to keep this change, resulting in the removal of the Types of Team Groups. @sapanathomas523  and team will add this back as part of a larger rollout.

en_us/shared/course_assets/pages.rst
mehaknasir: Hey, @nedbat Thank you for looking into this for me. I was confused if the hyperlink text is different from the title then we need to repeat it right?
I was under the impression if I use `:ref: Enable Wiki` then it will only show the title as a hyperlink but if we want to show a complete string like `11.3.2.3 `is extra here in the hyperlink then we need to specify exactly what we need to show as a hyperlink?

mehaknasir:
> From internal edX feedback, an additional requested change is in the Team Groups section
> 
> **Team Groups:** Each team _group_ is a grouping for multiple learner teams, **and each group has several configuration details and options.** To start, each of these team groups must be given a unique name, and optionally a description can also be set.
> 
> The bolded section can be removed. The italicized word should become group not groups.
> 
> FYI @sapanathomas523

This has been fixed

en_us/shared/course_assets/pages.rst
nedbat: Yes, it will only show the title, but why is the numbering useful?  By hardcoding it here, you've made it fragile: the numbering could change in the future, and this reference would be wrong.  And why does the user need the section numbering in the first place? This is a link: they will click it and be where they need to be.  The title alone will be fine.

mehaknasir: Yes, you are right. I never thought about that as it was written in the provided document as is.
@marcotuts should I remove numbering from all the hyperlinks as suggested by nat in the above discussion?

en_us/shared/course_features/teams/teams_setup.rst
mehaknasir: @saadyousafarbi this is the removed section we talking about

PR #1993, opened 2021-11-09
fix: bump remaining enterprise endpoints to v2

mattdrayer:
There were a few remaining references to the `v1` endpoint set in the Enterprise API documentation.  I have verified with Postman that the v2 endpoints are available in the API and updated the documentation, accordingly.

### Date Needed (optional)

Today.

### Reviewers

@akummins @georgebabey


akummins (approved):
LGTM!

PR #2001, opened 2022-01-03
fix: enhanced staff grader not in Maple

sarina:
Hi @pdpinch - was writing blog posts and noticed this feature hasn't yet landed (see active work on the feature branch: https://github.com/edx/edx-ora2/tree/feature/ora_staff_grader) - so I think we should remove the reference to ESG in the docs.

Let me know if you know more than I do about this. I couldn't really find any "esg" PRs in the repo that would indicate a big feature merge.


pdpinch:
Oh, is that what "ESG" means? 😁

In any case, that's a 👍 from me.

sarina:
Took me awhile to connect the dots! thanks

PR #2015, opened 2022-01-26
docs: Remove some old draft help center docs.

feanil:
These docs look like really old versions of the following two help center docs:

* https://support.edx.org/hc/en-us/articles/360004623354-When-I-share-my-program-record-what-information-is-shared-
* https://support.edx.org/hc/en-us/articles/360004623154-Viewing-and-sharing-your-program-record

I think this repo was used to get review and to get the initial draft written up and it no  longer needs to be here.
The help articles in the support center seem to be more up-to-date versions of this so I believe it's safe to clean this
up.

- [ ] Squash commits


Carlos-Muniz (approved):
This is a good cleanup. Maybe we should have a system for documentation readers to let us know if they find loose/superfluous/redundant information in an odd place, and a recommendation process on where that information should live, to help us with our organization and cleanup.

feanil:
@Carlos-Muniz that's a good idea, do you think the `edit on github` functionality that we can setup for docs is enough or were you thinking of something more?

Carlos-Muniz:
> @Carlos-Muniz that's a good idea, do you think the `edit on github` functionality that we can setup for docs is enough or were you thinking of something more?

@feanil 
I was thinking something more.
It would be good to have another link/button that would let someone detail their complaints, and automatically add that to issues.

PR #2019, opened 2022-02-03
docs: expanded flexible peer grading section

jansenk:
There was a misalignment in the docs and implementation for Flexible Peer Grading. The document stated that 
> earners who have received at least 30% of required peer-reviews who have waited longer than 7 days will be given their grade

but actually, we took 30 % and rounded down, so that statement was misleading.

I attempted to expand on the documentation to fix this issue and also provided some examples to make it clearer what the feature was actually doing.

[CR-4487](https://openedx.atlassian.net/browse/CR-4487)


marcotuts (approved):
this is great, thanks for improving the documentation!

PR #2025, opened 2022-02-25
docs: update bulk email documentation

justinhynes:
[MICROBA-1666]

Updates the bulk course email documentation to inform course teams of specific HTML content that is barred from use in messages authored from the Bulk Course Email tool. Previously, the message content authored by this tool was not being sanitized and thus we were allowing the storage of potentially dangerous HTML code in our systems. We will sanitize input from this tool before storing in the database and thus some HTML content may be "stripped" (converted to plaintext) when sent to users.

### DIFF
--- a/en_us/shared/manage_live_course/bulk_email.rst
+++ b/en_us/shared/manage_live_course/bulk_email.rst
@@ -173,17 +173,33 @@ When you compose an email message, you can use the following options.
 
 .. contents::
   :local:
-  :depth: 1
+  :depth: 2
 
 =======
 Styling
 =======
 
-Messages can include HTML styling, including text formatting and links. The
-email message editor offers the same formatting options as the Text component
-editor in Studio.
+Messages can include HTML styling, including text formatting and links.
 
 For more information, see :ref:`Working with Text Components`.
+
+-----------------------
+Restricted HTML Content
+-----------------------
+
+The Bulk Course Email tool restricts use of specific HTML elements from being
+used in messages. The following HTML elements may not be used in messages
+authored with this tool:
+
+* iframe
+* svg
+* form
+* script
+
+.. note: If a message contains restricted HTML content it will **not** be
+   stripped from the message. The HTML code will be converted to plain text. You
+   should send a test message to yourself first before sending to larger groups
+   to ensure message content is displayed as desired.
 
 ======
 Images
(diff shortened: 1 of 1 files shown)


en_us/shared/manage_live_course/bulk_email.rst
```diff
@@ -173,18 +173,34 @@ When you compose an email message, you can use the following options.
...
 =======
 
-Messages can include HTML styling, including text formatting and links. The
-email message editor offers the same formatting options as the Text component
-editor in Studio.
+Messages can include HTML styling, including text formatting and links.
 
 For more information, see :ref:`Working with Text Components`.
 
+-----------------------
```
justinhynes, line 186: I decided not to add a section of the items _allowed_ since this is configurable and can change.

The restricted content list isn't exhaustive but seemed like a good starting point of items I don't think we'd ever want to allow.

en_us/shared/manage_live_course/bulk_email.rst
```diff
@@ -173,18 +173,34 @@ When you compose an email message, you can use the following options.
...
+
+The Bulk Course Email tool restricts use of specific HTML elements from being
+used in messages. The following HTML elements may not be used in messages
+authored with this tool:
+
+* iframe
+* svg
+* form
+* script
```
Tj-Tracy, line 197: Should we link to the full approved list somehow or is this sufficient for 99% of cases?

Tj-Tracy (approved):
👍

en_us/shared/manage_live_course/bulk_email.rst
justinhynes: I passed on adding the full approved list as it can change at any time based on configuration.

hurtstotouchfire: Are there additional tags being filtered out by tinymce or is it basically the same set?

hurtstotouchfire (approved):
looks good to me! Had one non-blocking question.

en_us/shared/manage_live_course/bulk_email.rst
justinhynes: @hurtstotouchfire I tried to find a definitive list somewhere but could not locate it. I even tried to read through some of their source code but had difficulty grokking anything.

It seems like the tags filtered out by TinyMCE is pretty minimal by default and just contains the above items.

PR #2041, opened 2022-03-16
feat: Script to create .po files for Transifex Translations

### LINKED ISSUE #2040: feat: a script for creating translatable strings for Transifex

Carlos-Muniz:
The Documentation repo needed a consistent way of creating the .po files for a Transifex Project to add i18n to the Documentation. The `i18n.sh` script, written by @ghassanmas in https://github.com/openedx/edx-documentation/pull/1991, accomplishes the majority of the work needed to do this. 

Unfortunately, Transifex only allows uniquely named files and will overwrite non-unique names, so a second script `tx_config_fix.py` is used.  It renames the file names of the .po files that will be created in Transifex from the .pot files into unique names using the config section titles.

In addition, the requirements have been updated to reflect the new package requirements of `sphinx-intl` and `transifex-client`. And the Makefile has also been updated to include running of the `i18n.sh` script. 

Closes #2040


Carlos-Muniz:
@ghassanmas Do you mind reviewing this PR?

i18n.sh
```diff
@@ -0,0 +1,85 @@
...
+# already exists it will be overwritten!.
+# This script assumes sphinx-intl is already installed as well,
+# It can be installed by `pip install sphinx-intl`
+# For more information, about sphinx-intl please visit:
+# https://www.sphinx-doc.org/en/master/usage/advanced/intl.html
+
+
+# Check if this script is running by github, and if it's
+# missing the token then exit.
+if [ "$CI" = true ] && [ -z "$TX_TOKEN" ]; then
```
feanil:
```suggestion
if [ -z "$TX_TOKEN" ]; then
```
CI test seems not necessary and the message doesn't tell you anything about it.

i18n.sh
```diff
@@ -0,0 +1,85 @@
...
+# Root directory
+BASE_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+# Config file
+TX_CONFIG_FILE=$BASE_DIR/.tx/config
+
+# Success  exit message, wrapped in green color.
+EXIT_MESSAGE="\033[0;32mYou should now be able to push the files using tx client \n
+Typically this done by 'tx push -s', please note: you should have \n
+already installed tx client and configured it such that there is \n
+a project called 'open-edx-documentation-project' in the organization \n
```
feanil: You mention a specific project name here but then ask the user to provide their own project name. Maybe set this as the default name and allow the user the ability to override it?

i18n.sh
```diff
@@ -0,0 +1,85 @@
...
+    sphinx-intl update-txconfig-resources \
+    --pot-dir $BASE_DIR/$project/build/locale/ \
+    --locale-dir $BASE_DIR/$project/locales/ \
+    --transifex-project-name "$PROJECT_NAME"
+
+    python tx_config_fix.py "$project"
+
+done
+
+# Todo: Check for errors and set exit message accordingly
```
feanil: You should do this.  I often just use a `set -eoux pipefail`([Explanation](https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425))

tx_config_fix.py
```diff
@@ -0,0 +1,27 @@
+from configparser import ConfigParser
```
feanil, line 1: Add an explanation of what this script is doing as a docstring to the top of this file. Ideally with an example of what problem it's solving with a real example.

tx_config_fix.py
```diff
@@ -0,0 +1,27 @@
...
+
+import sys
+
+project_path = sys.argv[1]
```
feanil: I would recommend using `click` or `argparse` instead of parsing the command manually.  It is a lot more robust and less error prone.

tx_config_fix.py
```diff
@@ -0,0 +1,27 @@
...
+
+config_path = "./.tx/config"
+
+config = ConfigParser()
+config.read(config_path)
+
+
+for section_title in config.sections():
+    if config.has_option(section_title, 'source_file'):
+        source_file_path = config.get(section_title, 'source_file')
```
feanil: Not a huge problem here but in general you can save yourself a retrieval and be more pythonic by setting a `fallback` on the `get` call and then just having the conditional below be something like:
```
if source_file_path and source.startswith(project_path):
```

tx_config_fix.py
```diff
@@ -0,0 +1,27 @@
...
+        if source_file_path[:len(project_path)] == project_path:
+            split_section_title = section_title.split('.')
```
feanil: If you definitely need this to be 2 parts, you should [limit the number of splits](https://docs.python.org/3/library/stdtypes.html#str.split) to just 2 so you don't get weird behavior if there are more dots in the title.

tx_config_fix.py
```diff
@@ -0,0 +1,27 @@
...
+            split_section_title[1] = project_path_friendly + split_section_title[1]
```
feanil: NIT: maybe also add a delimiter between the friendly path and the file title for clarity when reading the config.

`test.en_us--open_edx_students--OpenSFD_certificates` instead of `test.en_us--open_edx_studentsOpenSFD_certificates`

i18n.sh
Carlos-Muniz: I removed this entire section, including the comment.

i18n.sh
Carlos-Muniz: PROJECT_NAME now has a default option, and the project name in EXIT_MESSAGE now reflects whatever the PROJECT_NAME is assigned as.

i18n.sh
Carlos-Muniz: I used `set -eou pipefail` as there were too many outputs to follow otherwise.

tx_config_fix.py
Carlos-Muniz: A docstring has been added with an example.

tx_config_fix.py
Carlos-Muniz: sys has replaced with click

tx_config_fix.py
Carlos-Muniz: get now has a fallback to be more pythonic

tx_config_fix.py
Carlos-Muniz: Another delimiter between the delimited project and the file name has been added.

feanil (approved):

PR #2047, opened 2022-04-01
docs: Remove IE from supported browsers

sarina:
See https://github.com/openedx/edx-platform/pull/30162

Per https://github.com/openedx/frontend-wg/issues/3 we no longer support IE 🎉 

Note that the [Learner Help Center for edX.org](https://support.edx.org/hc/en-us/articles/206211848-What-are-the-system-requirements-and-supported-browsers-on-edX-) indicates that IE is not supported but the [Open edX learner's guide](https://edx.readthedocs.io/projects/open-edx-learner-guide/en/latest/front_matter/browsers.html) does not, so this update fixes that.

@davidjoy

### DIFF
--- a/en_us/shared/browsers.rst
+++ b/en_us/shared/browsers.rst
@@ -14,7 +14,6 @@ Most current browsers will work on edX.org. For best performance, we recommend t
 We also support the latest versions of:
 
 * `Microsoft Edge`_
-* `Microsoft Internet Explorer`_
 * `Safari`_
 
 


pdpinch (approved):
Thank you!

sarina:
@pdpinch I _think_ I may have jumped the gun on this one. I'm waiting for @davidjoy or @adamstankiewicz to confirm/deny

adamstankiewicz:
@sarina Moving forward, it is correct that we are no longer supporting IE 11. Unfortunately, it turned out the code changes necessary to drop support in reality for older browsers was blocked by the ongoing Node 16 upgrade. So, even though the Learner Help Center article doesn't say we support IE 11, we technically still do, but only until the Node 16 upgrade is complete and we migrate to the shared `@edx/browserslist-config` across all the repos.

sarina:
Thanks @adamstankiewicz . What is your recommendation for this PR (this is documentation that Open edX users see)? Should we hold off until it's officially unsupported?

adamstankiewicz:
> Thanks @adamstankiewicz . What is your recommendation for this PR (this is documentation that Open edX users see)? Should we hold off until it's officially unsupported?

I'd say we can go ahead with merging this PR, even if it's not quite yet fully unsupported. The primary rationale being:
* With IE still listed, community members may spend effort making sure things are compatible, which is time not well spent since we no longer have the requirement.
* By removing IE here, should we get any support requests or bug reports related to IE specifically, we can point to this documentation to make it explicitly clear such requests/reports will not be prioritized at this time.
* Consistency with what's on the Learner Help Center article.

adamstankiewicz (approved):

sarina:
Would anyone mind giving a +1 to https://github.com/openedx/edx-platform/pull/30162 as well?

PR #2050, opened 2022-04-05
docs: Create ADR on Documentation Translation

Carlos-Muniz:
This PR adds a new directory for Architecture Decision Records under `docs/decisions/` as well as the first (0001) ADR. 0001 proposes the process by which the Documentation in this repo will be translated, integrated, and presented. 

The purpose of the format of this document is to elicit feedback from the greater community as Documentation affects everyone, and consistently translated Documentation can reach even more people than before.

### DIFF
--- a/decisions/0001-develop-documentation-translation-process.rst
+++ b/decisions/0001-develop-documentation-translation-process.rst
@@ -20,7 +20,7 @@ Answers to the questions
 1. When is the translation process triggered?
 =============================================
  
-Every six months, a newly named Open edX release is created, which includes the documentation. When this release is "cut", the translation process should also be triggered. So in the next case, where the Nutmeg version of the Open edX software is released, the translation process should also be triggered to translate the Nutmeg Documentation.
+Every six months, a newly named Open edX release is created, which includes the documentation. When this release is "cut", the translation process should also be triggered. For example, where the Nutmeg version of the Open edX software is released, the translation process should also be triggered to translate the Nutmeg Documentation.
 
 2. How is the translation process triggered?
 =============================================
--- a/decisions/0001-develop-documentation-translation-process.rst
+++ b/decisions/0001-develop-documentation-translation-process.rst
@@ -58,4 +58,7 @@ Not all of the documentation lives in the openedx/edx-documentation repository. 
 Location in GitHub
 ******************
 
-The documentation will remain in openedx/edx-documentation. The translations can be located in a new repository openedx/documentation-translations.
+The documentation will remain in openedx/edx-documentation. It time permits, engineering work will be done so that the translations can be located in a new repository named openedx/documentation-translations. If not, the translations will be co-located with the original strings as per the pattern in other Open edX repositories.
+
+Moving the translations to their own repository could be beneficial. Keeping the translations separate from the code that generates the documentation could dcrease repo clone/pull time. In addition, all translations could be kept in the same repo, and only the translations needed could be pulled from this repo. Lastly, when a new release of the documentation is cut, it can be separate from the translations that may be updating over a longer period of time, but still connected to a specfic version of the documentation.
+
--- a/decisions/0001-develop-documentation-translation-process.rst
+++ b/decisions/0001-develop-documentation-translation-process.rst
@@ -77,15 +77,15 @@ strings will look for identical strings from the Documentation in the Nutmeg
 release and automatically populate them.
 
 Upcoming work will automate the process of syncing files between GitHub and
-Transifex. Once a preset fraction of the strings are translated, a notice will
-be triggered to Transifex Admins that translations are ready. This will create
-a new Pull Request with these new translations. A maintainer of the
-edx-documentation repo will be asked for a review, and in combination with
-Transifex Admins, merged. In addition, this syncing process can be run
-regularly after the threshold is reached in order to increase the fraction of
-strings translated.
+Transifex via an established but new to us GitHub App. A GitHub App is an
+integration between two services, in this case, GitHub and Transifex. Once a
+preset fraction of the strings are translated, the GitHub App integrated with
+Transifex will create a new Pull Request and notify Transifex Admins by adding
+them for a review. Once reviewed, a Transifex Admin will then merge the Pull
+Request. In addition, this syncing process can be run regularly after the
+threshold is reached in order to increase the fraction of strings translated.
 
-4. How will the translated documents be integrated back into the documentation?
+1. How will the translated documents be integrated back into the documentation?
 ===============================================================================
 
 Translated documents will be kept in a newly created separate repository to
--- a/decisions/0001-develop-documentation-translation-process.rst
+++ b/decisions/0001-develop-documentation-translation-process.rst
@@ -17,6 +17,7 @@ In this ADR, we will describe:
  #. When the translation process is triggered
  #. How the translation process is triggered
  #. How the translation process works
+ #. How the documentation will stay updated
  #. How the translated documents will be integrated back into the documentation
 
 
@@ -51,15 +52,16 @@ that we receive the results we want. That is not to say that some parts of this
 process cannot be automated.
 
 Bringing translatable documentation material into Transifex:
- #. When a new named release is cut, a Github Pull Request should be created
-    that tags Github users that are also Transifex Admins. This GitHub PR will
-    contain the generated files to be uploaded to Transifex.
- #. To complete the process, the Transifex Admin will pull the PR and locally
-    run a script which uses their own credentials that uploads the documents to
-    Transifex.
- #. The Transifex Admin would then merge the PR as a record that the material
-    is on Transifex and ready to be translated by the greater open source
-    community.
+ #. A new named Open edX release branch of edx-documentation repo is made that
+    matches the "open-release/<release name>" pattern.
+ #. A github action will create a Pull Request containing generated files that
+    contain the strings to be translated via Transifex. A Transifex Admin will
+    be tagged as a reviewer to complete the PR.
+ #. The Transifex Admin will then locally run the script that places the
+    generated files into a properly named Project in Transifex.
+ #. Once the Transifex Project is ready for translators, the Transifex Admin
+    will merge the PR, creating a record that the material is on Transifex and
+    ready to be translated by the greater open source community.
 
 3. How does the translation process work?
 =============================================
@@ -85,7 +87,33 @@ them for a review. Once reviewed, a Transifex Admin will then merge the Pull
 Request. In addition, this syncing process can be run regularly after the
 threshold is reached in order to increase the fraction of strings translated.
 
-1. How will the translated documents be integrated back into the documentation?
+4. How will the documentation stay updated?
+===========================================
+
+The documentation for each new Open edX release is often updated while the
+software it documents is updated. Unfortunately, this means that as the
+documentation is finalized, the strings that need to be translated, will change
+as well. Luckily, Transifex already supplies this `functionality`_. Strings
+that are updated will need to be translated, but strings that stay the same
+will keep their translations.
+
+We will first upload documentation to Transifex any branch that matches the
+pattern we currently use. We will also set the GitHub integration to re-upload
+document strings with every merge to that branch, that way Translators will
+always have the latest strings. Translators will then translate the strings.
+And as the documentation is updated, translators can also translate those
+updated strings.
+
+Is this ideal? No. Translators may have to retranslate the contents of a
+document that was already translated before it was updated. It is a concession
+we have to make if we want updated software, and we want those updates
+documented, and we want that documentation translated. There will always be
+some inefficiencies in this pipeline as those three tasks cannot work perfectly
+in parallel since software updates continuously.
+
+.. _functionality: https://docs.transifex.com/projects/updating-content/
+
+5. How will the translated documents be integrated back into the documentation?
 ===============================================================================
 
 Currently, translated documents are placed back into the same repositories that
@@ -106,6 +134,7 @@ would be able to select the documentation for Lilac in Spanish via dropdown
 menus.
 
 .. _python documentation: https://docs.python.org/3/
+
 
 *********
 Rationale
--- a/decisions/0001-develop-documentation-translation-process.rst
+++ b/decisions/0001-develop-documentation-translation-process.rst
@@ -1,0 +1,61 @@
+Develop the Documentation Translation Process
+#############################################
+
+
+Description
+***********
+
+The edx-documentation repo contains important documentation written in English. Many users have a requirement for the important information found in this repository to be available in other languages. In order to provide the documentation in a variety of languages, a process must be designed and agreed upon to serve the greater open source community.
+
+In this ADR, we will describe: 
+ #. When the translation process is triggered
+ #. How the translation process is triggered
+ #. How the translation process works
+ #. How the translated documents will be integrated along with the rest of the documentation
+
+
+Answers to the questions
+************************
+
+1. When is the translation process triggered?
+=============================================
+ 
+Every six months, a newly named Open edX release is created, which includes the documentation. When this release is "cut", the translation process should also be triggered. So in the next case, where the Nutmeg version of the Open edX software is released, the translation process should also be triggered to translate the Nutmeg Documentation.
+
+2. How is the translation process triggered?
+=============================================
+An Admin account must add documents to a Transifex Project in order for them to be translated. Admin accounts have access to financial information and the ability to delete Transifex Projects. In order to keep this part of the process secure, we believe that a human should be involved in this process to assure that financial information stays in the right hands and so that an accidental merge will not delete Transifex projects. Currently, a GitHub bot account with Admin access to Transifex updates the translations for several repositories. This is the old and insecure way of doing it that neither Transifex nor GitHub recommend. We are retiring that bot account in preference to this secure and modern method. We believe that elevated permissions should not be granted to a bot or an app, and we should instead do a bit of the work manually to ensure that we receive the results we want. That is not to say that some parts of this process cannot be automated. 
+
+Bringing translatable documentation material into Transifex:
+ #. When a new named release is cut, a Github Pull Request should be created that tags Github users that are also Transifex Admins. This GitHub PR will contain the generated files to be uploaded to Transifex.
+ #. To complete the process, the Github User/Transifex Admin will pull the PR and locally run a script which uses their own credentials that uploads the documents to Transifex. 
+ #. The user would then merge the PR as a record that the material is on Transifex and ready to be translated by the greater open source community.
+
+
+3. How does the translation process work?
+=============================================
+Translators use Transifex to translate strings stripped from the documentation. These strings are kept in a Project in Transifex, where each string belongs to a uniquely named document, which is related to a document in the documentation. Projects can also be grouped into "Transifex Memory", so that similar strings can rely on previous translations to help inform future translations. Once uniquely named documents are uploaded to a Transifex Project within a pre-existing group, similarly worded strings will automatically be translated. Strings that need attention will be highlighted for users to focus on. 
+
+Upcoming work will automate the process of syncing files between GitHub and Transifex. Once a preset fraction of the strings are translated, a notice will be triggered to Transifex Admins that translations are ready. This will create a new Pull Request with these new translations. A maintainer of the edx-documentation repo will be asked for a review, and in combination with Transifex Admins, merged. In addition, this syncing process can be run regularly after the threshold is reached in order to increase the fraction of strings translated.
+
+
+4. How will the translated documents be integrated along with the rest of the documentation?
+============================================================================================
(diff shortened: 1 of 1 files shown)


decisions/0001-develop-documentation-translation-process.rst
```diff
@@ -0,0 +1,61 @@
...
+ #. How the translated documents will be integrated along with the rest of the documentation
+
+
+Answers to the questions
+************************
+
+1. When is the translation process triggered?
+=============================================
+ 
+Every six months, a newly named Open edX release is created, which includes the documentation. When this release is "cut", the translation process should also be triggered. So in the next case, where the Nutmeg version of the Open edX software is released, the translation process should also be triggered to translate the Nutmeg Documentation.
```
sarina: Nit: For those reading in the future, you may want to switch "So in the next case" with "For example,"

decisions/0001-develop-documentation-translation-process.rst
```diff
@@ -0,0 +1,61 @@
...
+
+Next Steps
+**********
+Not all of the documentation lives in the openedx/edx-documentation repository. In the future documentation will continue to be distributed amongst a number of repositories. The process described here will be used as a guide to bring Translations efficiently and consistently to all Open edX documentation repositories.
+
+
+Location in GitHub
+******************
+
+The documentation will remain in openedx/edx-documentation. The translations can be located in a new repository openedx/documentation-translations.
```
sarina: Question: Do we have the pattern of keeping translations in a separate repository? This is something I've wanted for awhile - particularly the ability to keep translations separate from the code (to decrease repo clone/pull time), and, crucially I think, have the ability to have _all_ translations in that repo, with an instance pulling only the translations they need. IIRC we don't have this and I don't know what's involved with doing this.

sarina (commented):
Question: This ADR does not mention how we would leverage Translation Memory. Specifically: say the Nutmeg documents are fully translated. Now we cut the Olive branch. Do we need to re-translate Olive strings from scratch, or is there a way to use the previous release's translations to help bootstrap?

decisions/0001-develop-documentation-translation-process.rst
Carlos-Muniz: That's a good note. I'll change it so it makes more sense for future readers.

decisions/0001-develop-documentation-translation-process.rst
Carlos-Muniz: We do not currently have that pattern, but this could change with the restructuring of the Documentation. There are a lot of advantages to keeping translations separate from the rest of the documentation. The one sticky point is that I'm not sure if ReadTheDocs supports the pulling of translations from a separate repo.

Carlos-Muniz:
> Question: This ADR does not mention how we would leverage Translation Memory. Specifically: say the Nutmeg documents are fully translated. Now we cut the Olive branch. Do we need to re-translate Olive strings from scratch, or is there a way to use the previous release's translations to help bootstrap?

We should be able to use the translated Nutmeg documents in Transifex Memory to speed up the time it takes to translate the documents in the Olive branch. This can be done by placing all documentation projects in the same group, so they have the same Transifex Memory. We may still have to translate strings in Olive that are not 100% identical to strings that have been translated before.

Thanks for this question @sarina, I will update the ADR to contain this information.

decisions/0001-develop-documentation-translation-process.rst
sarina: Yeah, that's what I thought. I might be more clear in this sentence, something like, "If time, engineering work will be done so that the translations would be located in a new repository. If not, the translations will be co-located with the original strings as per the pattern in other openedx repositories."

Possibly too it'd be good to expand on why this work would be justified/what the advantages are.

Carlos-Muniz: Definitely. I'll add a couple sentences to that effect.

decisions/0001-develop-documentation-translation-process.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Bringing translatable documentation material into Transifex:
+ #. When a new named release is cut, a Github Pull Request should be created that tags Github users that are also Transifex Admins. This GitHub PR will contain the generated files to be uploaded to Transifex.
```
sarina: Do we want to dip our toe just a little deeper into technical details? IE you say, 

> When a new named release is cut, a Github Pull Request should be created that tags Github users that are also Transifex Admins.

Who makes this PR? How do they remember to do so? Is it maybe part of a script that BTR runs when creating the release?

decisions/0001-develop-documentation-translation-process.rst
```diff
@@ -0,0 +1,64 @@
...
+ #. To complete the process, the Github User/Transifex Admin will pull the PR and locally run a script which uses their own credentials that uploads the documents to Transifex.
+ #. The user would then merge the PR as a record that the material is on Transifex and ready to be translated by the greater open source community.
```
sarina: I think in 2 & 3 I'd just use "Transifex Admin" oven "user" for clarity. But take/leave that suggestion.

decisions/0001-develop-documentation-translation-process.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Upcoming work will automate the process of syncing files between GitHub and Transifex. Once a preset fraction of the strings are translated, a notice will be triggered to Transifex Admins that translations are ready. This will create a new Pull Request with these new translations. A maintainer of the edx-documentation repo will be asked for a review, and in combination with Transifex Admins, merged. In addition, this syncing process can be run regularly after the threshold is reached in order to increase the fraction of strings translated.
```
sarina: Again dipping our toes into implementation, is it worth elaborating just a tad on what the upcoming work is? I don't remember if you're using GH Apps or not, but I could see something like:

Upcoming work will automate the process of syncing files between GitHub and Transifex via a (new? pre-existing?) GitHub App. A GitHub App is an integration between two services, in this case, GitHub and Transifex.  Once a preset fraction of the strings are translated, the GitHub App will create a new Pull Request with these new translations.

I'm actually a bit unsure what you meant by

> a notice will be triggered to Transifex Admins that translations are ready. This will create a new Pull Request with these new translations.

What is the form of the notice? The way this reads is that Transifex Admins will receive the notice and make a manual PR.

decisions/0001-develop-documentation-translation-process.rst
```diff
@@ -0,0 +1,64 @@
...
+
+
+4. How will the translated documents be integrated along with the rest of the documentation?
+============================================================================================
+Translated documents will be kept in a newly created separate repository to avoid having to recut a named Open edX release everytime translations are added or updated. During the building process of the documents, it will check for the existence of these translated documents, and build with them if they exist. Translations will be available both by locale and by selection as is currently supported by Sphinx and Read the Docs. So for example, while the default option for document readers in the US locale will be English, they will also be able to select other translated documents.
```
sarina: > Translated documents will be kept in a newly created separate repository

Ah, I missed this was included in scope of work, but in the "Location in GitHub" section it's been changed to say that's add on work if time. Let's make sure those are in sync. I still think you should do the first round work of simply getting the integration with the translations in-repo, then tackle this as a Phase 2 to make sure you have incremental deliverables.

Have you done research on how this might be accomplished?

decisions/0001-develop-documentation-translation-process.rst
```diff
@@ -0,0 +1,64 @@
...
+
+
+Location in GitHub
+******************
+
+The documentation will remain in openedx/edx-documentation. It time permits, engineering work will be done so that the translations can be located in a new repository named openedx/documentation-translations. If not, the translations will be co-located with the original strings as per the pattern in other Open edX repositories.
+
+Moving the translations to their own repository could be beneficial. Keeping the translations separate from the code that generates the documentation could dcrease repo clone/pull time. In addition, all translations could be kept in the same repo, and only the translations needed could be pulled from this repo. Lastly, when a new release of the documentation is cut, it can be separate from the translations that may be updating over a longer period of time, but still connected to a specfic version of the documentation.
```
sarina:
```suggestion
Moving the translations to their own repository could be beneficial. Keeping the translations separate from the code that generates the documentation could decrease repo clone/pull time. In addition, all translations could be kept in the same repo, and only the translations needed could be pulled from this repo. Lastly, when a new release of the documentation is cut, it can be separate from the translations that may be updating over a longer period of time, but still connected to a specific version of the documentation.
```

sarina (commented):
Hey this is looking great!

One style comment, you're inconsistent in your spacing before and after headers. I'd recommend following the RST style guide here: https://documentation-style-guide-sphinx.readthedocs.io/en/latest/style-guide.html#headings

decisions/0001-develop-documentation-translation-process.rst
ghassanmas: +1 on this I also think "Location in GitHub" better be detailed out intially, since the decision on this is going to affect most if not all phases of i18n of the doc.. more detials in my comment below

ghassanmas:
@cmuniz Thank you for work!
I have went through the PR and before writing my comments, I opted to doing a bit of discovery work/literature review[1] and frankly its taking a fair amount of time. 

I just have a quick question for you or for anyone in community involved in the documenation as well: 

*Do we currenly support releasing doc for various version? When I browse the doc on readthedocs I can only choose the latest but I guess this would need to change when publising?* 

Also another reason why I wanted to do more discovery work is that all the steps/phases on the ADR (in my opinion) would be affected by our decison on where/how the doc is going to be placed i.e. "Location in GitHub", precisely its going to affect: 

- The PR which create translation files
- The syncing of the doc /  the location of the doc 
-  The building of the doc 

[1]: I am currently looking at how the doc of sphinx and for django are lied out, both projects support i18n of the doc, however only django seems to support viewing the doc for multiple versions.

decisions/0001-develop-documentation-translation-process.rst
Carlos-Muniz: This makes sense. I'll change it to Transifex Admin

decisions/0001-develop-documentation-translation-process.rst
Carlos-Muniz: fixed!

ghassanmas:
@nedbat  I am pinging you because you are one who created this change on [github](https://github.com/openedx/edx-documentation/commit/73deb294cac70b58971437d6c45200fb5683b176): if I am not mistaken this change imply that only for proejct prefixed open-edx we are publishing doc branches for named released. 
However what I can't find is, do you we have a hidden webhook that once we pushes open-release/*.master  it will trigger RTD to rebuild it? or is there some sort of configuration that is hidden? e.g configured via RTD admin panel?

I am asking this because I just want to know what is mechnism used for publishing multiple versions, which to my understanding can be accomplished either via RTD, or through a sphinx-extension. And then I wonder if we are going to keep relying on RTD to build the doc, then this would imply we have to rely on it for generating the i18n?.

ned:
I think you mean @nedbat, I'm a different ned 😅

decisions/0001-develop-documentation-translation-process.rst
Carlos-Muniz: If possible, this should be a github action. Unfortunately, at the moment, edx-platform is not really published in a way with a lot of finality. As far as I know, it does a soft publication into a branch. Even if we had an action for the creation of a new branch that follows the scheme open-release/*.master, this would still probably not be the best way of doing this.

I guess the details pertaining to this question depend on the process of cutting a new release.

decisions/0001-develop-documentation-translation-process.rst
Carlos-Muniz: Ah, thats not my intention.
The PR will be made by the Github integration. And Transifex Admins should be notified that the PR was made by being tagged as reviewers to approve and merge the PR.

decisions/0001-develop-documentation-translation-process.rst
Carlos-Muniz: I will update the text to mention tackling this work as a phase 2.

As for research on how the translations and documentation versions can be selected, the [python documentation](https://docs.python.org/3/) which is built with sphinx, offers both of these. We can just copy what they do directly from the [Doc subdirectory of the Python Repo](https://github.com/python/cpython/tree/main/Doc).

sarina: thanks Carlos!

Carlos-Muniz:
Hey @ghassanmas I am answering your review questions inline:

> Do we currently support releasing doc for various version?

Open edX does ~~not~~ currently support various versions of the documentation on ReadTheDocs. This means that you will ~~not~~ be able to find both the Lilac and Maple versions of the documentation rendered in the same place on ReadTheDocs. You can still find it in this repo. 

~~I believe this is already part of the plan to upgrade the documentation.~~

> all the steps/phases on the ADR (in my opinion) would be affected by our decision on where/how the doc is going to be placed i.e. "Location in GitHub", precisely its going to affect:
>The PR which create translation files
>The syncing of the doc / the location of the doc
>The building of the doc

That's correct: all of these steps will be effected. But we can still make these changes now before the big `edx-documentation` upgrade, and when the upgrade happens, we can then tweak the scripts and automations to look in the right places. And given that this is an important upgrade for the greater Open edX community, it would be good for the future of the documentation to keep the structure described here in mind when designing the future structure.

> I am currently looking at how the doc of sphinx and for django are lied out, both projects support i18n of the doc, however only django seems to support viewing the doc for multiple versions.

We are not planning on moving away from RTD/Sphinx for our documentation needs. Luckily, Sphinx does support viewing multiple versions of the documentation and multiple languages as well. Check out the [Python Documentation](https://docs.python.org/3/) for an example.

> I am asking this because I just want to know what is mechanism used for publishing multiple versions

Python Documentation uses a custom Sphinx Theme called [python-docs-theme](https://github.com/python/python-docs-theme/blob/1201e82bc2f5e28d365428ad73fde4700b1f52f6/python_docs_theme/layout.html#L20) that allows users to switch the documentation between languages and switch the documentation between versions

ghassanmas:
@Carlos-Muniz 
Yeah I was just looking over python doc per you reference, and it seems they have built their own way of handeling multiple versions and lanagaues from scractch[1], however there key differences from our method:
- They don't use Transfiex instaed they rely open-source tools developed by Mozila [1]
- They have their own tools/scripts to build the docs i.e. they don't rely on read the docs for building nor publishing. 

In contrast they do pull translation from various repo, they have a dedicated repo for each lang, for example this the repo for [french](https://github.com/python/python-docs-fr), and when building they pull the content from all repos or spesfic (depends on the build arg)[4],  

> Open edX does not currently support various versions of the documentation on ReadTheDocs. This means that you will not be able to find both the Lilac and Maple versions of the documentation rendered in the same place on ReadTheDocs. But you can still find it in this repo.

After further checking it seems this is activated but only for projects that starts/prefixed with open-edx, check [3] I was able to select lilac and koa, but not maple this what I was curious to ask @nedbat about.  

> Python Documentation uses a custom Sphinx Theme called [python-docs-theme](https://github.com/python/python-docs-theme/blob/1201e82bc2f5e28d365428ad73fde4700b1f52f6/python_docs_theme/layout.html#L20) that allows users to switch the documentation between languages and switch the documentation between versions 

I don't think they do it through the theme, instead they rely on their custom scripts repo, hence check the `/templates` in [2]. 


[1] Check PEP-0545 https://peps.python.org/pep-0545/ 
[2] The scripe are in this repo: https://github.com/python/docsbuild-scripts
[3] Open-edx-student doc at RTD https://edx.readthedocs.io/projects/open-edx-learner-guide/en/latest/
[4] https://github.com/python/docsbuild-scripts/blob/main/build_docs.py#L337-L357

ghassanmas:
[Correction] I mentioned earlier that python doc doesn't rely on Transfiex, but that is actually incorrect, as I was refrencing [PEP545](https://peps.python.org/pep-0545/#translation-tools), here is what they exactly say about the translation tooling:

> [Translation tools](https://peps.python.org/pep-0545/#translation-tools)
Most of the translation work is actually done on Transifex [[15]](https://peps.python.org/pep-0545/#id40).

> Other tools may be used later like https://pontoon.mozilla.org/ and http://zanata.org/. 

So I wrongfully conclude that their alternative option as their choosen tool. May be just useful to know there are open source or free alternative to TX if cost become an issue at somepoint. 

Sorry for the inconvenience folks! :sweat_smile:.

Carlos-Muniz:
@ghassanmas Thanks for doing all this discovery work.
I'll update the ADR with your findings and our discussion.

decisions/0001-develop-documentation-translation-process.rst
Carlos-Muniz: The translation of the documentation of each new named Open edX release should begin when the branch is formed. Unfortunately, this means that as the documentation is finalized, the strings that need to be translated, will change as well. Luckily, Transifex already supplies this [functionality](https://docs.transifex.com/projects/updating-content/). Strings that are updated will need to be translated, but strings that stay the same will keep their translations.

We first upload documentation to Transifex any branch that matches the pattern we currently use. And we set the GitHub integration to re-upload document strings with every merge to that branch. Translators will then translate the document. And as the documentation is updated, translators can also translate those updated strings.

Is this ideal? No. Translators may have to retranslate the contents of a document that was already translated before it was updated. It is a concession we have to make if we want updated software, and we want those updates documented, and we want that documentation translated. There will always be some inefficiencies in this pipeline as those three tasks cannot work perfectly in parallel since software updates continuously.

decisions/0001-develop-documentation-translation-process.rst
```diff
@@ -0,0 +1,176 @@
...
+Admin access to Transifex updates the translations for several repositories.
+This is the old and insecure way of doing it that neither Transifex nor GitHub
+recommend. We are retiring that bot account in preference to this secure and
+modern method. We believe that elevated permissions should not be granted to a
+bot or an app, and we should instead do a bit of the work manually to ensure
+that we receive the results we want. That is not to say that some parts of this
+process cannot be automated.
+
+Bringing translatable documentation material into Transifex:
+ #. A new named Open edX release branch of edx-documentation repo is made that
```
sarina: perhaps "is made by an automated script" (if that is right... I believe this process isn't manual)

decisions/0001-develop-documentation-translation-process.rst
```diff
@@ -0,0 +1,176 @@
...
+translated and is in Transifex Memory, when the Olive release is cut, the
+strings will look for identical strings from the Documentation in the Nutmeg
+release and automatically populate them.
+
+Upcoming work will automate the process of syncing files between GitHub and
+Transifex via an established but new to us GitHub App. A GitHub App is an
+integration between two services, in this case, GitHub and Transifex. Once a
+preset fraction of the strings are translated, the GitHub App integrated with
+Transifex will create a new Pull Request and notify Transifex Admins by adding
+them for a review. Once reviewed, a Transifex Admin will then merge the Pull
```
sarina: Why can't this PR be auto-merged?

decisions/0001-develop-documentation-translation-process.rst
```diff
@@ -0,0 +1,176 @@
...
+merge will not delete Transifex projects. Currently, a GitHub bot account with
+Admin access to Transifex updates the translations for several repositories.
+This is the old and insecure way of doing it that neither Transifex nor GitHub
+recommend. We are retiring that bot account in preference to this secure and
+modern method. We believe that elevated permissions should not be granted to a
+bot or an app, and we should instead do a bit of the work manually to ensure
+that we receive the results we want. That is not to say that some parts of this
+process cannot be automated.
+
+Bringing translatable documentation material into Transifex:
```
sarina, line 54:
```suggestion
Bringing translatable documentation material into Transifex:

```

sarina (approved):

decisions/0001-develop-documentation-translation-process.rst
Carlos-Muniz: It can be, but then very little of this process is supervised by a human. And I thought that was one of the things we wanted.

sarina: In other repos I believe the PR to fold translations back into the repo is auto-merged. What would human eyes add here?

The only thing that I feel is very necessary to involve a human are those things requiring administrative privileges.

decisions/0001-develop-documentation-translation-process.rst
Carlos-Muniz: Added + link

decisions/0001-develop-documentation-translation-process.rst
Carlos-Muniz: ~~Added new-line before lists~~
This turns the list below it into a quote, which I don't want.

decisions/0001-develop-documentation-translation-process.rst
Carlos-Muniz: I didn't realize that translations were auto-merged in other repos.
I guess human eyes unfamiliar with the translated and reviewed strings would not add any other value here.

I'll change this.

PR #2092, opened 2022-11-10
Move branch and tag info from the developer docs

feanil:
The branch and tag info is useful but it was kept in the edx-developer-docs
which are about to be deprecated.  We move this info closer to the other
release related docs.

We'll want to further refactor this into the new docs site in the future but
just co-locating with other usefule release information for now.

### DIFF
--- a/en_us/open_edx_release_notes/source/index.rst
+++ b/en_us/open_edx_release_notes/source/index.rst
@@ -11,6 +11,7 @@ The *Open edX Platform Release Notes* provide information about releases, migrat
 
     front_matter/index
     os_documentation
+    named_release_branches_and_tags
     olive
     nutmeg
     maple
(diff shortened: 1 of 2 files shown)


kdmccormick (approved):
Nice. I built it locally and found that this page actually fits fairly naturally into the existing documentation. Can you link to the DEPR issue for edx-developer-docs in your commit message?

FYI @mtyaka  and @nedbat -- The edx-developer-docs repo is being deprecated. It had a "Named Release Branches and Tags" page that we kept updated as part of the release process, so Feanil is moving that page into edx-documentation until we can find a new home for it at docs.openedx.org.

@feanil , do you think you'll have edx-developer-docs decommissioned by the Olive release a month from now? Just need to know whether BTR should update this page both here and in the old location, or whether we can forget about the old location.

feanil:
@kdmccormick The deprecation is accepted as of Friday so I think I'll have it all the way cleaned up by olive.  I've got an edit for the BTR Release checklist that I'll save on the wiki when I land this as well so that you just need to update one place. 

P.S. One caveat to this is that I am currently in the process of moving all the release notes to the new docs site so this may move again. TBD if that will land before the Olive cut but I'll do my best to update the BTR docs when that happens as well.

kdmccormick:
That sounds great, thanks a bunch Feanil.

PR #2104, opened 2023-02-08
build: Change how we run `make upgrade`.

feanil:
Previously we use to run all the complie commands from the top-level
directory. We change it so that we run the command from within the
`requirements` directory so that the output paths match how pip-compile
is being run by dependabot.

This is an experiment to see if we can replace our own github actions
with Dependabot for python upgrades.


nedbat:
This seems like an OK change, but I don't understand how dependabot fits into it?

feanil:
Dependabot now seems to be able to run `pip-compile` on `*.in` files and produce the relevant `*.txt` file.  However it does this in the same directory as the `.in` file so the output is slightly different than when we do it from a different directory.  I want to try to use dependabot here but also enable our `make upgrade` tooling to see if they catch different things or if they are mostly the same.  However, without this change there will be some comments that will keep changing back an forth like they did in https://github.com/openedx/edx-documentation/pull/2096

nedbat (approved):

PR #2106, opened 2023-02-09
feanil/redirect release notes

feanil:
Now that the release notes are [live on the new docs site](https://docs.openedx.org/en/latest/community/release_notes/index.html): Update the old site to just redirect everything to the new site.

I also removed all the content from here so it's clear here the source of truth is.

Right now this site still gets a lot of traffic so we should not take it down anytime soon:
image


feanil:
The PR build of the notes is here: https://edx--2106.org.readthedocs.build/projects/open-edx-release-notes/en/2106/

And a second sample redirect:
https://edx--2106.org.readthedocs.build/projects/open-edx-release-notes/en/2106/olive.html

run_tests.sh
```diff
@@ -79,7 +79,7 @@ for project in "${projects[@]}"; do
     # -w writes warnings and errors to the specified file in
```
sarina, line 79: I'd remove the `-j4` reference at L78

sarina:
Is my understanding correct that this change strips the content, instead pointing it to the relevant content on docs.openedx.org?

run_tests.sh
feanil: It's fine up there because that is being run just to generate the inventory files and we don't care if it produces warnings there.  There's a bunch of cross-inventory referencing so we have to build twice, once to generate the inventories and a second time to make sure that the inventories are all cross-linked correctly.

feanil:
Correct, the new release notes are live, this points the old site to the new one including all sub-pages to their equivalent in the new deployment.

sarina (approved):

PR #2110, opened 2023-02-21
build: Add python upgrade job.

feanil:
Add a description of your changes with links to any relevant material.

### DIFF
--- a/.github/workflows/upgrade-python-requirements.yml
+++ b/.github/workflows/upgrade-python-requirements.yml
@@ -1,0 +1,30 @@
+name: Upgrade Python Requirements
+
+on:
+  schedule:
+    - cron: "0 0 * * 1"
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: "Target branch against which to create requirements PR"
+        required: true
+        # If copying this template manually, you must provide your default branch name
+        # in quotes, such as 'master'
+        default: 'master'
+
+jobs:
+  call-upgrade-python-requirements-workflow:
+    uses: openedx/.github/.github/workflows/upgrade-python-requirements.yml@master
+    with:
+    # If copying manually, also provide your default branch name in quotes here
+      branch: ${{ github.event.inputs.branch || 'master' }}
+      # optional parameters below; fill in if you'd like github or email notifications
+      # user_reviewers: ""
+      # team_reviewers: ""
+      # email_address: ""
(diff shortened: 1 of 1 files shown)


.github/workflows/upgrade-python-requirements.yml
```diff
@@ -0,0 +1,30 @@
...
+        # If copying this template manually, you must provide your default branch name
+        # in quotes, such as 'master'
+        default: 'master'
+
+jobs:
+  call-upgrade-python-requirements-workflow:
+    uses: openedx/.github/.github/workflows/upgrade-python-requirements.yml@master
+    with:
+    # If copying manually, also provide your default branch name in quotes here
+      branch: ${{ github.event.inputs.branch || 'master' }}
```
nedbat, line 20: The input has a default above (line 13).  Do we need the default here also?

.github/workflows/upgrade-python-requirements.yml
```diff
@@ -0,0 +1,30 @@
+name: Upgrade Python Requirements
+
+on:
+  schedule:
+    - cron: "0 0 * * 1"
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: "Target branch against which to create requirements PR"
```
nedbat, line 9:
```suggestion
        description: "Target branch for requirements PR"
```

.github/workflows/upgrade-python-requirements.yml
feanil: This is a copy from the template in `.github` so I'm not entirely sure. Do you want me to simplify this and see if it will work?  We can always make a follow-up PR to add this back if we find a good reason for it.

nedbat: I was afraid it was a copy from a template, but didn't go to find the template.  I understand not wanting to diverge from the template, it just seemed like a bit of code we didn't need.

nedbat (approved):
Your choice what to do with my comments.

.github/workflows/upgrade-python-requirements.yml
feanil: Ah, I figured it out,  if this is triggered by `workflow_dispatch` then `github.events.inputs.branch` will be set but that value will not be set if it's triggered via `schedule`.  I think that's why it's like this.

PR #2112, opened 2023-02-28
docs: add configuration documentation on how to enable returning extended profile metadata through the User API

justinhynes:
[APER-2285]

This PR adds documentation around configuration of the extended profile feature of the User API. The User API of edx-platform allows the storage of additional metadata in a user's profile instance without needing to extend the underlying model. However, this data won't be returned unless the Open edX system is configured to return the data. This feature has existed for some time but I couldn't find any documentation no how to configure it.

### DIFF
--- a/en_us/install_operations/source/configuration/deprecated_add_course_talk_widget.rst
+++ b/en_us/install_operations/source/configuration/deprecated_add_course_talk_widget.rst
@@ -1,15 +1,15 @@
 .. _Add CourseTalk (Deprecated):
 
-################################################
-DEPRECATED: This Is Unavailable in Lilac onwards
-################################################
-
 #############################################
 Adding the CourseTalk Widget
 #############################################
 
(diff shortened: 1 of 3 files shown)


en_us/install_operations/source/configuration/deprecated_add_course_talk_widget.rst
```diff
@@ -1,16 +1,15 @@
 .. _Add CourseTalk (Deprecated):
 
-################################################
-DEPRECATED: This Is Unavailable in Lilac onwards
-################################################
-
```
justinhynes, line 6: I'm fixing this while I'm in here. Converting this to a note in the main body of this document. Otherwise, this adds a separate section to the outline that is confusing:
image

MaxFrank13 (commented):
Looks good! Thanks for updating the documentation

MaxFrank13 (approved):

PR #2158, opened 2023-07-03
docs: Add maintainership information.

feanil:


feanil:
@nedbat I realized that I'm watching this repo a lot and making sure it stays working so may as well update the maintainership docs.  Do you want to be co-maintainer?

nedbat (approved):

PR #2163, opened 2023-07-17
docs: flex peer ora course override

jansenk:
Add documentation for ORA card with link to flexible override documentation


leangseu-edx (changes requested):
Can you make a file `ora_settings.rst` instead? We are expecting this to multiple setting maybe? And we are expecting that file https://github.com/openedx/edx-platform/blob/8780e606f70790e226115f797887ed6343a1e9b2/cms/envs/common.py#L2700.

jansenk:
> Can you make a file `ora_settings.rst` instead? We are expecting this to multiple setting maybe?

I don't want things to be any more spread out than they are already, but I did adjust the formatting of this page. Any additional settings will also go here.

> And we are expecting that file https://github.com/openedx/edx-platform/blob/8780e606f70790e226115f797887ed6343a1e9b2/cms/envs/common.py#L2700.

Good point. I'll make sure to make that change before we release.

mattcarter (approved):
I'm approving, with the idea that the agreed to changes will get done.

leangseu-edx (approved):

jansenk:
https://github.com/openedx/edx-platform/pull/32791

PR #2206, opened 2023-11-01
Move XBlock Tutorial to XBlock repo

sarina:
@feanil - depends on https://github.com/openedx/XBlock/pull/686


en_us/xblock-tutorial/conf.py
```diff
@@ -12,3 +12,9 @@
 
 # remove directory when content is first added to it, and add to index
 exclude_patterns = ['reusable/*']
+
+
+redirects = {
+    ## TODO
+    "*": "https://docs.openedx.org/en/latest/developers/references/developer_guide/$source.html",
```
feanil: The xblock docs will be available under this url: https://docs.openedx.org/projects/xblock/en/latest/ so I think we'll want the redirect to be somewhere under there.

sarina: They're at https://edx.readthedocs.io/projects/xblock/en/latest/xblock-tutorial/index.html so I assume this shoud be `https://edx.readthedocs.io/projects/xblock/en/latest/xblock-tutorial/$source.html`?

feanil: I've updated this so they're also published under the docs.openedx.org domain here: https://docs.openedx.org/projects/xblock/en/latest/xblock-tutorial/index.html 

The docs.openedx.org domain version is the same content but at a more correct location.

sarina: Thanks @feanil added that in

feanil (approved):

sarina:
@feanil do you know why the two checks aren

sarina:
@feanil do you know why the two checks aren't running (docs/readthedocs.org:edx-developer-guide and open-edx-release-notes)? We moved both of those guides anyway so shouldn't we remove these checks?

feanil:
Looks like those two checks are required in GitHub but maybe no longer setup.  I agree that they should just be removed from the set of required checks.

sarina:
Hmm. I tried removing them from `run_tests.sh` but that didn't make the build icon go away (still says it's "expected"). I'm not sure how to disable the check otherwise.

feanil:
Oh yea the update is in the github settings but it looks like this is actually related to a new change on the RTD side, I'm fixing that now.

sarina:
Should I back this commit out? https://github.com/openedx/edx-documentation/pull/2206/commits/140ab26f2e987a35610167c17b92a0942f3d257e

feanil:
Yea, you should remove that commit, I think it's still good to test the redirect builds are working as expected and should be pretty fast.

PR #2254, opened 2024-05-15
Update standard workflow files.

feanil:
This PR was created automatically by [the `repo_checks` tool](https://github.com/openedx/repo-tools/tree/master/edx_repo_tools/repo_checks).


feanil:
We don't want this on docs.openedx.org since there are a lot of non technical contributors and we don't want to make things hader for them.

PR #2261, opened 2024-05-23
build: Fix the xblock-tutorial build so that redirects work.

feanil:
When the code was removed, things got re-arranged into a weird state that made it not possible to compile the RST
anymore.  This change restores the makefile and adds enough scaffolding so that the build will succeed and point to the
new documentation.


sarina (approved):
Ah, ok probably my bad in the yaml file. For the Makefile, will something like this be needed every time we migrate a filetree?

feanil:
I'm not sure the makefile will be needed for RTD builds but it is needed to locally build so I think leaving it in there is fine.

PR #2302, opened 2024-11-15
docs: Remove badger docs that got missed.

feanil:
We dropped badgr as a part of https://github.com/openedx/edx-platform/issues/31541


sarina:
Curious why we'd remove somethnig from previous release notes when it's just no longer implemented? What about a note at the top of the page indicating that support was dropped in `<RELEASE NAME>`?

feanil:
I added a note to the list to acknowledge that the feature was there but dropped later and that's why there are no docs on the latest version of the docs site.

sarina:
I don't agree that just removing outdated docs here is the right thing, since this was available until Palm. But, your call I guess.

feanil:
While the code on our side existed till Palm, Badgr has not existed as a service for a while.  Having this in the latest doc here confused someone into thinking it was still available and especially since the service it integrates with doesn't exist anymore, I don't see a lot of value in the docs.

sarina:
Ah! Gotcha, I thought this was simply an integration we removed at some point, but it's a service that no longer exists at all. This makes sense to me.

sarina (approved):

PR #2305, opened 2024-11-27
docs: clarify ora waiting details

jansenk:
- The waiting step details link is always clickable when there is a peer step. The docs said otherwise
- Included a clarifying note about why learners who have been given an override can still show up in the table


nsprenkle (approved):
Good clarification :shipit:

PR #2318, opened 2025-02-11
build: Update intersphinx mapping

sarina:
@feanil - I saw you did this for the [XBlock docs](https://github.com/openedx/edx-documentation/commit/d230bdcf0c93c954341fbbe562d89184eeff4210), so I figured I should do it for other docs that have moved. However I'm not sure what else I need to do to effect redirects. Note we are also currently in the process of moving the OLX docs (https://github.com/openedx/docs.openedx.org/pull/819)

### DIFF
--- a/shared/conf.py
+++ b/shared/conf.py
@@ -209,19 +209,19 @@ def ism_location(dir_name):
         return None
 
 intersphinx_mapping = {
-    "opencoursestaff" : (openedx_rtd_url("open-edx-building-and-running-a-course"), ism_location("open_edx_course_authors")),
+    "opencoursestaff" : ("https://docs.openedx.org/en/latest/", None),
     "data" : (edx_rtd_url("devdata"), ism_location("data")),
     "partnercoursestaff": (edx_rtd_url("edx-partner-course-staff"), ism_location("course_authors")),
     "insights" : (edx_rtd_url("edx-insights"), None),
     "xblockapi" : ("https://docs.openedx.org/projects/xblock/en/latest/", None),
-    "xblocktutorial" : (edx_rtd_url("xblock-tutorial"), ism_location("xblock-tutorial")),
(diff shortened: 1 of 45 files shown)


feanil (commented):
You shouldn't point to the index.html but to the `en/latest` url because intersphinx builds a URL on top of that to find the inventory files that are generated as a part of docs builds.

sarina:
@feanil okay thanks, how do I point at https://docs.openedx.org/en/latest/educators/index.html? Would I be able to do https://docs.openedx.org/en/latest/educators - or can I only do https://docs.openedx.org/en/latest?

sarina:
So, I'm getting a million build failures related to missing labels, and I'm going to assume a part of that is that when we moved docs, we changed a number of the labels....

sarina:
@feanil this is ready to go. I'll open up separate PRs to figure out the redirects for the open edx course staff docs and eventually the OLX docs once https://github.com/openedx/docs.openedx.org/pull/819 is merged

shared/conf.py
```diff
@@ -209,19 +209,19 @@ def ism_location(dir_name):
         return None
 
 intersphinx_mapping = {
-    "opencoursestaff" : (openedx_rtd_url("open-edx-building-and-running-a-course"), ism_location("open_edx_course_authors")),
+    "opencoursestaff" : ("https://docs.openedx.org/en/latest/", None),
```
feanil, line 212: FYI, this is because you set the second parameter to `None` that you had to update all the prefixes.  Not something you need to fix but for future reference.

feanil (approved):

sarina:
Relates to https://github.com/openedx/public-engineering/issues/324

shared/conf.py
sarina: ah gotcha, ty!

PR #2321, opened 2025-02-13
Redirect OLX URLs to docs.openedx.org and remove migrated OLX docs

sarina:
The OLX docs have been moved (ref: https://github.com/openedx/docs.openedx.org/pull/819) so redirect OLX URLs to the new site.

I'm not doing a wildcard, because we're going to be rewriting/moving docs based on Diataxis, so the redirect will point to the OLX docs landing page.

### Reviewers

@feanil @jswope00


sarina:
Relates to openedx/public-engineering#324

feanil (approved):

PR #2332, opened 2025-03-05
Add a custom 404 page via sphinx-notfound-page extension

sarina:
- **build: add sphinx-notfound-page extension**
- **build: Add a custom 404 page to the docs**

In the course of trying to add this 404 page, I encountered an error in pydata_sphinx_theme where the toctree parsing is expecting an actual toctree rather than an empty page tree. The sphinx-notfound-page extension made this error be thrown when it wasn't being thrown before because it adds an empty toctree (https://github.com/readthedocs/sphinx-notfound-page/blob/main/notfound/extension.py#L147-L148). I worked around it by adding `dummy.rst` pages to hidden toctrees in all app directories that didn't have subpages, but I'm open to hearing other ways of working around it.

### DIFF
--- a/en_us/contribute/source/dummy.rst
+++ b/en_us/contribute/source/dummy.rst
@@ -1,0 +1,4 @@
+Dummy page
+##########
+
+.. pydata_sphinx_theme wants there to be a populated toctree, not an empty index page, so populate the toctree :(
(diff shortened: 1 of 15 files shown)


feanil (approved):
Seems weird but these pages are never visible since these projects are all getting redirected right? So I think it's fine.

sarina:
I'm going to go ahead and merge this as-is so we can get the custom 404 messages out, but I've made an upstream issue to figure out if there's a better way to resolve this - https://github.com/readthedocs/sphinx-notfound-page/issues/253

